PR #1, opened 2015-09-11
Member Resource

mikejritter:
Adds a member resource to the dpn-server API.

Updates include:
- Creating a member model/controller/presenter
- Creating db migrations for members
- Adding relation to bags
- Updating bags_controller to include members
- Adding test for new resources and updating

These changes should be added to the dpn-rest-wiki repository as well.

### DIFF
--- a/app/controllers/api_v1/bags_controller.rb
+++ b/app/controllers/api_v1/bags_controller.rb
@@ -94,7 +94,7 @@ class ApiV1::BagsController < ApplicationController
   def create
     expected_params = [:bag_type, :rights, :interpretive, :uuid,
       :local_id, :size, :version, :ingest_node, :admin_node,
-      :replicating_nodes, :first_version_uuid, :fixities
+      :replicating_nodes, :first_version_uuid, :fixities, :member
     ]
 
     unless expected_params.all? {|param| params.has_key?(param)}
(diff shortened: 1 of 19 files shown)


malakai97:
1. The association name should just be bags, i.e. a member has bags, not member bags.  This makes sense as when you're coding you'll always have the member when you retrieve this association.  I.e. it's obvious that they're that member's bags.  See below:
   
   ``` ruby
   some_member = Member.find_by_uuid(uuid)
   some_members_bags = some_member.bags
   ```
2. Bags should be presented with the member's uuid, not the member's name.  We can't guarantee the name is unique.  This seems to be done both ways, as the controller expects a member's uuid, and the routes also use uuid, not name.  
3. Does your institution require you to add any license info?  I don't know what happens for files I created that you edited, but it seems that for the ones you created something might be required.  
- [ ] Change association to member has_many bags, instead of has_many member_bags.
- [ ] Change to always serialize bags as displaying the member uuid.
- [ ] Add any copyright notices/headers required by your institution.

mikejritter:
1 and 2 are easy changes. A bit more work for 2, just to make sure anything which relied on the name is updated as well.

3 is a bit tricky as I don't actually know and the person I can ask is on vacation until the 28th. Looking through our docs it looks like my institution has started to put GPL 2.1 as the license for most of the open source software we write, but I don't know of any requirements.

malakai97:
No reason to block on #3.  That information can be added later.  If it was a big issue for your institution, you'd know.

PR #5, opened 2015-09-30
Test data

diamondap:
I merged test-data with master two days ago and resolved conflicts. I merged again this morning, and all spec tests are passing, so it should be a clean, easy merge.

I'm still having problems with integrations tests, but I'll file those issues separately.


malakai97:
Changes manually split by type and merged onto master.

PR #11, opened 2016-02-08
Clean config

malakai97:
# This pull request redefines how configuration is done

Some additional steps are required in order to adopt the changes.  Instructions follow.  The README has also been updated and contains detailed, general instructions.
## I used environment variables before and don't want to change anything

You should pull the changes, then execute `bundle exec rake config:deprecated`.  This will generate configuration files that behave exactly as before, and will add dotenv to your Gemfile.local
## I did not like way we did things before but I still want to use environment variables

Environment variable use is still supported, but you will have to explicitly setup your database.yml and secrets.yml files to read from the environment.  Consult the updated README and the .example files.
## I don't want to use environment variables

The entire config is now contained in database.yml, secrets.yml, and Gemfile.local.  Consult the updated README and the .example files.
# What happened to capistrano?

I've removed capistrano from the project because it's deployment configuration that does not belong in the application.  It's still possible to deploy an application with a capistrano configuration outside of the repo, and in fact this gives each organization a lot more control over their deployment.


malakai97:
Keep this PR around for the instructions for now.

PR #19, opened 2016-07-06
Add dpn_cops for rubocop support

dazza-codes:
- Adds the `dpn_cops` gem via git HTTPS
- Configures the `dpn_cops` using the `.rubocop_todo.yml`
  - many of the offenses can be fixed by `rubocop --auto-correct` (not applied in this PR)


dazza-codes:
Discarding this PR - discovered that code-climate cannot support including cop definitions from a gem.

PR #20, opened 2016-07-06
NodeAdapter - resolve storage region/type id by name

dazza-codes:
This PR slightly changes syntax in the node adapter, i.e.
https://github.com/dpn-admin/dpn-server/blob/master/app/adapters/node_adapter.rb#L31-L36

### DIFF
--- a/app/adapters/node_adapter.rb
+++ b/app/adapters/node_adapter.rb
@@ -40,8 +40,8 @@ class NodeAdapter < ::AbstractAdapter
       storage_type_id: nil
     }
     if storage.respond_to? :has_key?
-      region = StorageRegion.find_by_name(storage[:region])
-      type   = StorageType.find_by_name(storage[:type])
+      region = StorageRegion.find_by(name: storage[:region])
+      type   = StorageType.find_by(name: storage[:type])
       result[:storage_region_id] = region ? region.id : nil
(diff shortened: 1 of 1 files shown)


malakai97:
Were you only seeing this behavior when using the cluster fixtures?  If so, that could explain the buggy behavior you were seeing.

`app/models/storage_{region,type}.rb` uses `app/models/concerns/lowercased.rb` to force downcasing of values for `name`.  It also overrides some of the query methods to downcase the search term before searching.  

Fixtures override this behavior, so uppercase characters may be stored to the database.  As well, the `Lowercased` concern does not override `::find_by`.  These two facts together indicate to me that the fixture is storing an uppercased value, and the overridden `#find_by_name` method is failing to find it because it's looking for a lowercased value.

Assuming I'm correct, the fix is to:
1. Now: Change the fixture's values to be lowercase.
2. Future: Update `Lowercased` to override the model's `::find_by` behavior.  This could be nontrivial as we don't want to break its other uses.
3. Consider whether or not fixtures are the best way to load data in the impersonate/cluster config.

dazza-codes:
Why is lower case enforced by the concern?

malakai97:
"Other" != "other"
"chron" != "Chron"
etc.  

The concern enforces saving only as lowercase, and downcases the value passed to #find_by_name(value)

PR #21, opened 2016-07-08
Fixup/models update

malakai97:
Mostly improving encapsulation and paying down some technical debt.


malakai97:
Should be against develop, not master!

PR #25, opened 2016-07-26
Replication bag

### LINKED ISSUE #23: replication_transfer GET response could include `bag_id`?
The replication_transfer GET response includes a `uuid` value, which seems to correspond to a bag UUID.  A replication_transfer POST invokes validation that requires a `bag_id` value.
- https://github.com/dpn-admin/dpn-server/blob/master/app/models/replication_transfer.rb#L106
- https://github.com/dpn-admin/dpn-server/blob/master/app/models/replication_transfer.rb#L123

So, perhaps the GET response `uuid` could be `bag_id` instead (or copied to `bag_id`)?  If it were, it could be easier to directly create/update the replication_transfer using the GET response in a corresponding POST/PUT request.  The related code seems to be in
https://github.com/dpn-admin/dpn-server/blob/master/app/adapters/replication_transfer_adapter.rb#L15

For example, a GET response for some fixity data looks like this:

``` ruby
{:replication_id=>"20000000-0000-4000-a000-000000000007",
 :fixity_nonce=>nil,
 :fixity_value=>nil,
 :fixity_accept=>false,
 :bag_valid=>true,
 :status=>"requested",
 :link=>"/data/src/dlss/dpn/dpn-server/test/fixtures/integration/testbags/00000000-0000-4000-a000-000000000002.tar",
 :created_at=>"2015-09-15T19:38:31Z",
 :updated_at=>"2015-09-15T19:38:31Z",
 :uuid=>"00000000-0000-4000-a000-000000000002",
 :fixity_algorithm=>"sha256",
 :protocol=>"rsync",
 :from_node=>"chron",
 :to_node=>"aptrust"}
```

This response is derived from the fixity data at
https://github.com/dpn-admin/dpn-server/blob/master/test/fixtures/integration/chron/replication_transfers.yml#L15-L27

Notice how it no longer contains the `bag: chron01` value and it has added some values, including the `uuid` that corresponds to a `bag_id` for the POST validations.

dazza-codes:
A cherry-pick of commits that fix https://github.com/dpn-admin/dpn-server/issues/23 into master.

### DIFF
--- a/app/adapters/replication_transfer_adapter.rb
+++ b/app/adapters/replication_transfer_adapter.rb
@@ -12,7 +12,7 @@ class ReplicationTransferAdapter < ::AbstractAdapter
   map_simple :status, :status
   map_simple :link, :link
 
-  map_belongs_to :bag, :uuid
+  map_belongs_to :bag, :bag, sub_method: :uuid
   map_belongs_to :fixity_alg, :fixity_algorithm, sub_method: :name
   map_belongs_to :protocol, :protocol, sub_method: :name
   map_belongs_to :from_node, :from_node, model_class: Node, sub_method: :namespace
--- a/app/adapters/restore_transfer_adapter.rb
+++ b/app/adapters/restore_transfer_adapter.rb
@@ -9,7 +9,7 @@ class RestoreTransferAdapter < ::AbstractAdapter
   map_simple :status, :status
   map_simple :link, :link
 
-  map_belongs_to :bag, :uuid
+  map_belongs_to :bag, :bag, sub_method: :uuid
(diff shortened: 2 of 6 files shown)


malakai97:
Assuming I understand this correctly, and despite this change being correct behavior (thanks!), I'd rather not break clients that rely on this incorrect behavior ahead of v2.  Does that sound reasonable?

dazza-codes:
Given the proximity of v2 (merge develop -> master), these cherry-picks can wait.  Given that they are cherry-picks, it should make no difference to the merge.  But, if the master is redeployed and some clients expect the `uuid` instead of the `bag`, your right, they will break.

PR #31, opened 2016-08-11
dpn-242_bug fix

smutniak:
fixing dpn-242 bug by changing default datetime in db/schema.rb

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -197,7 +197,7 @@ ActiveRecord::Schema.define(version: 20160805224340) do
 
   create_table "run_times", force: :cascade do |t|
     t.string   "name",                                          null: false
-    t.datetime "last_success", default: '1970-01-01 00:00:00Z', null: false
+    t.datetime "last_success", default: '1970-01-01 00:00:00', null: false
   end
 
   add_index "run_times", ["name"], name: "index_run_times_on_name", unique: true


malakai97:
There are a couple of things going on here.
1. The db/schema.rb is generated by rails, primarily by running rake db:migrate.  So this change is likely to be overwritten.
2. This file will be different depending on which database adapter (sqlite, mysql, postgres, etc) ActiveRecord has been configured to use.  This is despite rails devs claiming schema.rb is agnostic.  The version committed in the repo is for sqlite.  You'll notice that it doesn't have an foreign keys, despite the migrations creating them.  This is because sqlite doesn't have foreign key support.  

One we we could go about finding these types of errors before they hit production is configuring travis to build the application once with sqlite, once with mysql, etc.

It may be that the best way to fix this is to remove db/schema.rb from the repo.  That would force users to generate their own with the database adapter they've configured.  However, there's little way to guarantee that a correct schema.rb is generated--it seems like we'd be asking for trouble.  

tl;dr: This fix isn't going to address the underlying issue, and I'm not aware of a good strategy for addressing the underlying issue at this time.

malakai97:
Closing this as this isn't a bug.  See my previous comment.

smutniak:
Makes sense.  I'll remove the dpn-242_bug branch.

PR #36, opened 2016-08-22
Gemfile - relax ruby version

dazza-codes:
Allow any ruby `2.3.x` version; e.g. travis is configured to run `2.3.1` (at present).


dazza-codes:
Forget it, the spec allows `2.3.1`, as is.  I just ran into a problem because my default ruby was `2.2.5`.

PR #47, opened 2016-08-24
fixed impersonate configs

dazza-codes:
This is a fix for some conflicts in the impersonate configs between `config/impersonate.yml` and the new `config/dpn.yml` for api-v2 (in `develop`).  On a clean clone and checkout of the branch of this PR, I was able to run:

```
git clone git@github.com:dpn-admin/dpn-server.git
cd dpn-server/
git checkout impersonate-config-fix
rbenv local 2.3.1
# might need `gem install bundle` before:
bundle install
bundle exec rake config
bundle exec ./script/setup_cluster.rb
bundle exec ./script/run_cluster.rb -f
```

The motivation for this fix to the configs was that the prior `RAILS_ENV=impersonate` was not specific enough for each node and the `local_namespace` was not set correctly because of conflicts between the prior `config/environments/impersonate.rb` and `config/initializers/dpn.rb`.  This PR resolves all these conflicts (after some significant analysis and testing).

Prior to this fix, we got a strange value for the `local_namespace` setting, because it was using a 'hathi' value from the `config/dpn.yml` file for every one of the cluster nodes, e.g.

```
$ RAILS_ENV=impersonate_sdr rails c
Loading impersonate_sdr environment (Rails 4.2.7.1)
irb(main):001:0> Rails.application.config.local_namespace
=> "hathi"
```

Using the fixes in this PR, we get the correct `local_namespace`, e.g.

```
$ RAILS_ENV=impersonate_sdr rails c
Loading impersonate_sdr environment (Rails 4.2.7.1)
irb(main):001:0> Rails.application.config.local_namespace
=> "sdr"
```


config/environments/impersonate.rb
```diff
@@ -1,29 +0,0 @@
-require Rails.root.join("config/environments/development")
```
dazza-codes, line 1: All the settings in this file are now handled by `config/environment/development`, because it pulls in data from `config/secrets`.  So the script in `script/setup_impersonate` can simply copy the development into a node-specific environment.  Hence, the `.gitignore` can now ignore all of the impersonate environment files and this file is removed.  This file was actually serving as a wrapper for node-specific environment files.

config/impersonate.yml
```diff
@@ -1,38 +0,0 @@
-defaults: &defaults
```
dazza-codes, line 1: Most of the content in this file is now in `config/secrets` and `config/dpn`; see the `lib/template/...` files for details.

script/run_cluster.rb
```diff
@@ -47,7 +47,7 @@
 nodes = %w(aptrust chron hathi sdr tdr)
 
 def common_assignments(node)
-  "RAILS_ENV=impersonate IMPERSONATE=#{node} DATABASE_URL=sqlite3:db/impersonate_#{node}.sqlite3"
+  "RAILS_ENV=impersonate_#{node} IMPERSONATE=#{node} DATABASE_URL=sqlite3:db/impersonate_#{node}.sqlite3"
 end
```
dazza-codes, line 51: Note that the RAILS_ENV is now a node-specific environment so that it gets the correct content from `config/secrets` and `config/dpn`.

smutniak:
I just tested this and it worked fine.  I'll vote to merge.

PR #50, opened 2016-08-30
MySQL specs for api-v2 (develop)

dazza-codes:
A PR to compare changes required to replace sqlite with MySQL.  A useful branch for evaluating db migrations and specs using MySQL.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,24 @@
 language: ruby
 rvm:
   - 2.3.1
-branches:
-  only:
-    - master
-    - develop
-script: bundle exec rspec --tag ~broken_in_ci
-install:
-  - bundle install --jobs=3 --retry=3
-  - bundle exec sqlite3 --version
-  - bundle exec rake config
+
 addons:
   code_climate:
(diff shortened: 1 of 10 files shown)


malakai97:
This is a good first step towards supporting a matrix of backends in travis ci builds.  We will first need to do it in a more robust way that cleanly supports simultaneous builds.  

Let's keep this up to date with develop.

sqlite will remain the default backend for dev+test.  It's important that the project can be rapidly cloned and the tests run in the way that is expected for ruby and rails open-source projects.

dazza-codes:
This is the argument for merging this with develop:
- ActiveRecord has some inconsistent behavior on sqlite vs. mysql
- 4/5 DPN nodes are using MySQL in production
- MySQL setup is trivial given a linux-distro installation and using `db/mysql_init.sql` in this branch
- Developing with MySQL should avoid last-minute problems with deployment
- When using MySQL, specs can be run without `--tag ~broken_in_ci`

dazza-codes:
Closing this PR to move the branch from `mysql-specs` to `api-v2-mysql-specs`.

PR #53, opened 2016-08-31
[Testing] MySQL specs for api-v2 (develop)

dazza-codes:
This branch can be rebased on `api-v2` (which is currently rebased on `develop`, but may be later rebased on `master`).  This branch should allow testing migrations using MySQL instead of SQLite.


dazza-codes:
Closing this to use the `develop-mysql` branch instead, for more frequent checks.

PR #54, opened 2016-08-31
db:migrate tweaks that worked for SDR deployment

dazza-codes:
These two commits split up a couple of migrations into more atomic units.  This helped to resolve some deployment failures in the migrations on MySQL for `dpn-dev.stanford.edu`.  One significant change in these commits is the use of a custom name for the foreign key constraints, because our deployment platform was using the same FK identifier for the constraints and refusing to create the FK for fixity_checks on nodes.

### DIFF
--- a/db/migrate/20160429211034_add_timestamps_to_message_digests.rb
+++ b/db/migrate/20160429211034_add_timestamps_to_message_digests.rb
@@ -1,0 +1,26 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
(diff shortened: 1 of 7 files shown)


db/migrate/20160429211034_add_timestamps_to_message_digests.rb
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
+class AddTimestampsToMessageDigests < ActiveRecord::Migration
+  def change
+    add_timestamps(:message_digests, null: false)
+  end
```
dazza-codes, line 25: This adds both `created_at` and `updated_at` so active record can apply all it's default behaviors.

PR #57, opened 2016-09-01
Fix bad initialization in Client::Sync::Job#perform

### LINKED ISSUE #56: Sync jobs fail to initialize
The common sync job is calling `classify`, when it should be calling `constantize`.  This is an easy fix, bu the tests should have caught this.

malakai97:
This resolves #56

### DIFF
--- a/spec/jobs/client/sync/job_spec.rb
+++ b/spec/jobs/client/sync/job_spec.rb
@@ -13,33 +13,35 @@ describe Client::Sync::Job do
   model_hash = {foo: 1, bar: 2}
 
   describe "#perform" do
-    before(:each) do
-      @job = Client::Sync::Job.new
-      allow(@job).to receive(:last_success_manager)
-      allow(@job).to receive(:remote_client)
-      allow(@job).to receive(:query_builder)
-      allow(@job).to receive(:sync)
+    let(:job) do
+      job = Client::Sync::Job.new
+      allow(job).to receive(:last_success_manager)
+      allow(job).to receive(:remote_client)
+      allow(job).to receive(:query_builder)
+      allow(job).to receive(:sync)
+      job
     end
-    subject { @job.perform("some_name", "some_namespace", "Fixnum", "String", "Time") }
+
+    subject { job.perform("some_name", "some_namespace", "Fixnum", "String", "Time") }
 
     it "creates a last_success_manager" do
-      expect(@job).to receive(:last_success_manager).with("some_name", "some_namespace")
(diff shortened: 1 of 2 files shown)


spec/jobs/client/sync/job_spec.rb
```diff
@@ -11,65 +11,100 @@
...
 
+  describe "#perform" do
+    before(:each) do
+      @job = Client::Sync::Job.new
+      allow(@job).to receive(:last_success_manager)
+      allow(@job).to receive(:remote_client)
+      allow(@job).to receive(:query_builder)
+      allow(@job).to receive(:sync)
+    end
+    subject { @job.perform("some_name", "some_namespace", "Fixnum", "String", "Time") }
```
dazza-codes: Looks OK, although a `let` idiom might be preferred, e.g.

``` ruby
let(:job) { described_class.new }
```

Then use `job` instead of `@job` in the `before(:each)` block (and remove the `@job` init).  Place the `let` above the `before` block, so it's defined when reading the before block; although rspec execution may not require that, it's easier to read.

dazza-codes:
One small suggestion on using a `let` clause instead of an instance variable for `job`.  Not sure I fully understand the intricacies of the stubs and mocks.  Maybe they were too aggressive and obscured the bug in the prior revision of these specs.

malakai97:
This class could probably be designed better, but ActiveJob places some restrictions on what kind of arguments perform can accept, which causes it to be awkward.

I haven't really seen a way for the `let` functionality to work when one part of setup must refer to another part, as the first part is unitialized during setup.

dazza-codes:
You can use a block with `let`, so if you always want to stub those methods as no-ops, it would be

``` ruby
let(:job) do
  j = described_class.new
  # stub methods as no-op
  allow(j).to receive(:last_success_manager)
  allow(j).to receive(:remote_client)
  allow(j).to receive(:query_builder)
  allow(j).to receive(:sync)
  j
end
```

Also consider `let!` when you definitely want it to create something every time.  Most, if not all, of the instance variables in these specs could be `let` statements.  Try running `gem install rubocop` (outside Gemfile inclusion, but in your ruby version gems), and run `rubocop {spec_file_path}` to get useful tips on rspec conventions (it might require another gem for rspec, not sure).  Also look at the auto-correct feature of rubocop to see what changes it would make.

For now, this seems to be working and this is the style your comfortable with, so it's not a blocker.  I just want to encourage consideration of the rubocop conventions that make it easier for open-source projects to adopt and use a consistent community style.   Having consistent styles makes a world of difference for easier reading and review of code.

PR #68, opened 2016-09-09
Minimal changes extracted from swagger-ui work

dazza-codes:
This PR comprises some minimal changes that were introduced while working on swagger-ui enhancement.  These change can be reviewed and accepted separately, to isolate this PR review from the core changes required in the swagger-ui branch.  Once these are accepted, the swagger-ui PR can proceed and should be easier to review.

The gem changes in this PR are mainly reorganizing the asset gems into a group.  While working on swagger-ui, the asset gems were updated, using a clean rails 4.2 app as a guideline on the gems required for asset compilation.  There are a couple of laptop-dev additions for using rubocop and the thin web server.


malakai97:
This looks fine in general. Two questions:
1. Can we squash these commits into one or two commits
2. What functionality is `thin` providing over Rails's built-in application server?

dazza-codes:
1.  The commits are in a state where they can be individually reverted, if need be.  Squashing them might interfere with that.
2.  Thin is a faster server than WeBrick, so tests should run quicker when run against it.  This is more important for integration tests that run against the cluster.  A more complete answer is found elsewhere, e.g. http://www.madebymarket.com/blog/dev/ruby-web-benchmark-report.html
   
   2.1  While working on swagger-ui, it was helpful to test it on HTTPS and thin supports that.

malakai97:
According to the benchmarks on that page Webrick is about 90% as fast as Thin.  Additionally, both can support thousands of requests per second, dwarfing our use case.

Webrick [supports HTTPS](http://ruby-doc.org/stdlib-2.3.1/libdoc/webrick/rdoc/WEBrick.html#module-WEBrick-label-HTTPS) out of the box.

dazza-codes:
OK, removed thin.

PR #69, opened 2016-09-13
dpn-server will now automatically check fixity, request storage

### LINKED ISSUE #63: Bag transfer - StoreRequested set true for good fixity value from sync-client
_From @darrenleeweber on September 2, 2016 18:56_

From #61 
- [ ] dpn-server should set StoreRequested to true on replication transfers when the to_node returns a good fixity value.

_Copied from original issue: sul-dlss/dpn-server#7_

malakai97:
fixes #63 

This solution uses an after_update callback.  There are some other such callbacks already present on the ReplicationTransfer model.  This functionality is fine for now, but we'll want to move away from callbacks in the future, e.g. with the command pattern.  Such a change does would not affect the API.

### DIFF
--- a/app/models/replication_transfer.rb
+++ b/app/models/replication_transfer.rb
@@ -25,7 +25,7 @@ class ReplicationTransfer < ActiveRecord::Base
     unless cancelled
       transaction do
         update!(cancelled: true, cancel_reason: reason)
-        bag_man_request.cancel!(reason)
+        bag_man_request&.cancel!(reason)
       end
     end
   end
@@ -46,6 +46,7 @@ class ReplicationTransfer < ActiveRecord::Base
   after_create :add_request_if_needed
   after_update :preserve_bag_if_needed
   after_update :add_replicating_node_if_needed
+  after_update :request_storage_if_needed
 
 
   ### Static Validations
@@ -99,6 +100,25 @@ class ReplicationTransfer < ActiveRecord::Base
 
   private
 
+  # True if we created this request
+  def we_requested?
+    from_node && from_node.namespace == Rails.configuration.local_namespace
+  end
+
+
+  # If we are the from_node, check fixity
+  # after it is supplied.
+  def request_storage_if_needed
+    if we_requested? && fixity_value && !cancelled? && !store_requested
+      if fixity_value == bag.message_digests.where(fixity_alg_id: fixity_alg_id).first.value
+        update!(store_requested: true)
+      else
+        self.cancel!("fixity_reject")
+      end
+    end
+  end
+
+
   # Cancelled records are read-only.
   def no_changes_once_cancelled
     if changed? && cancelled_was == true
@@ -138,7 +158,7 @@ class ReplicationTransfer < ActiveRecord::Base
   # Adds the replicating node to the bag when it is marked
   # as stored.  Only needed if we are the from_node.
   def add_replicating_node_if_needed
-    if stored_changed?(from: false, to: true) && from_node&.namespace == Rails.configuration.local_namespace
+    if we_requested? && stored_changed?(from: false, to: true)
       bag.replicating_nodes << to_node
     end
   end
--- a/app/models/replication_transfer.rb
+++ b/app/models/replication_transfer.rb
@@ -28,6 +28,7 @@ class ReplicationTransfer < ActiveRecord::Base
         bag_man_request&.cancel!(reason)
       end
     end
+    return true
   end
 
 
@@ -44,9 +45,6 @@ class ReplicationTransfer < ActiveRecord::Base
 
   ### Callbacks
   after_create :add_request_if_needed
-  after_update :preserve_bag_if_needed
-  after_update :add_replicating_node_if_needed
-  after_update :request_storage_if_needed
 
 
   ### Static Validations
@@ -100,25 +98,6 @@ class ReplicationTransfer < ActiveRecord::Base
 
   private
 
-  # True if we created this request
-  def we_requested?
-    from_node && from_node.namespace == Rails.configuration.local_namespace
-  end
-
-
-  # If we are the from_node, check fixity
-  # after it is supplied.
-  def request_storage_if_needed
-    if we_requested? && fixity_value && !cancelled? && !store_requested
-      if fixity_value == bag.message_digests.where(fixity_alg_id: fixity_alg_id).first.value
-        update!(store_requested: true)
-      else
-        self.cancel!("fixity_reject")
-      end
-    end
-  end
-
-
   # Cancelled records are read-only.
   def no_changes_once_cancelled
     if changed? && cancelled_was == true
@@ -141,25 +120,6 @@ class ReplicationTransfer < ActiveRecord::Base
     if to_node&.namespace == Rails.configuration.local_namespace
       self.bag_man_request = BagManRequest.create!( source_location: link, cancelled: false)
       self.bag_man_request.begin!
-    end
-  end
-
-
-  # Instructs the bag manager request to preserve the bag
-  # when storage is requested, typically by the from_node.
-  # Does nothing if there is no bag man request, typically
-  # if we are not the to_node.
-  def preserve_bag_if_needed
-    if store_requested_changed?(from: false, to: true)
-      bag_man_request&.okay_to_preserve!
-    end
-  end
-
-  # Adds the replicating node to the bag when it is marked
-  # as stored.  Only needed if we are the from_node.
-  def add_replicating_node_if_needed
-    if we_requested? && stored_changed?(from: false, to: true)
-      bag.replicating_nodes << to_node
     end
   end
 
--- a/app/services/replication_transfer_updater.rb
+++ b/app/services/replication_transfer_updater.rb
@@ -1,0 +1,142 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
+class ReplicationTransferUpdater
+
+  class << self
+    def update(record, params)
+      build_update(record, params).update
+    end
+
+    def build_update(record, params)
+      case
+      when CancelUpdate.matching_update?(record, params)
+        CancelUpdate.new(record, params)
+      when FixityValueUpdate.matching_update?(record, params)
+        FixityValueUpdate.new(record, params)
+      when StoreRequestedUpdate.matching_update?(record, params)
+        StoreRequestedUpdate.new(record, params)
+      when StoredUpdate.matching_update?(record, params)
+        StoredUpdate.new(record, params)
+      else
+        ReplicationTransferUpdate.new(record, params)
+      end
+    end
+
+  end
+
+
+  # A generic update
+  class ReplicationTransferUpdate
+
+    # @param [ReplicationTransfer] record
+    # @param [HashWithIndifferentAccess] params The params hash of the update
+    def initialize(record, params)
+      @record = record
+      @params = params
+    end
+    def update
+      record.update(params)
+    end
+    def we_requested?
+      record.from_node&.namespace == Rails.configuration.local_namespace
+    end
+    def we_replicating?
+      record.to_node&.namespace == Rails.configuration.local_namespace
+    end
+
+    private
+    attr_reader :record, :params
+  end
+
+
+  # An update in which the request is cancelling the record
+  class CancelUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.cancelled == false && params[:cancelled] == true
+    end
+    def update
+      record.cancel!(params[:cancel_reason])
+    end
+  end
+
+
+  # An update in which the request sets fixity_value
+  class FixityValueUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.fixity_value.nil? && params[:fixity_value] != nil
+    end
+
+    def update
+      result = record.update(params)
+      if result && we_requested?
+        request_storage
+      end
+      return result
+    end
+
+    private
+
+    def request_storage
+      if params[:fixity_value] == correct_fixity
+        record.update(store_requested: true)
+      else
+        record.cancel!("fixity_reject")
+      end
+    end
+
+    def correct_fixity
+      record.bag.message_digests.where(fixity_alg_id: record.fixity_alg_id).first.value
+    end
+  end
+
+
+  # An update in which the request sets store_requested
+  class StoreRequestedUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.store_requested == false && params[:store_requested] == true
+    end
+
+    def update
+      result = record.update(params)
+      if result && we_replicating?
+        preserve_bag
+      end
+      return result
+    end
+
+    private
+
+    def preserve_bag
+      record.bag_man_request&.okay_to_preserve!
+    end
+
+  end
+
+
+  # An update in which the request sets stored
+  class StoredUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.stored == false && params[:stored] == true
+    end
+
+    def update
+      result = record.update(params)
+      if result && we_requested?
+        add_replicating_node
+      end
+      return result
+    end
+
+    private
+
+    def add_replicating_node
+      record.bag.replicating_nodes << record.to_node
+    end
+
+  end
+
+
+end
--- a/spec/models/replication_transfer_spec.rb
+++ b/spec/models/replication_transfer_spec.rb
@@ -35,6 +35,36 @@ describe ReplicationTransfer, type: :model do
       expect(r.update(stored: true)).to be true
       expect(r.bag.replicating_nodes).to include r.to_node
     end
+
+    context "when fixity_value set" do
+      let(:correct_fixity) { "98734723942304820348203840238402983409283408234" }
+      let(:fixity_alg) { Fabricate(:fixity_alg) }
+      let(:bag) {
+        bag = Fabricate(:bag_without_digests)
+        bag.message_digests << Fabricate.build(:message_digest, fixity_alg: fixity_alg, value: correct_fixity)
+        bag.save!
+        bag
+      }
+      let(:transfer) { Fabricate(:replication_transfer,
+        fixity_alg: fixity_alg,
+        bag: bag,
+        from_node: Fabricate(:local_node))
+      }
+      it "sets store_requested->true when correct" do
+        transfer.update!(fixity_value: correct_fixity)
+        expect(transfer.reload.store_requested).to be true
+      end
+      it "cancels with fixity_reject when incorrect" do
+        transfer.update!(fixity_value: "123908102831028301820398120398102830129830")
+        expect(transfer.reload.cancelled).to be true
+        expect(transfer.reload.cancel_reason).to eql("fixity_reject")
+      end
+      it "does not set store_requested when incorrect" do
+        transfer.update!(fixity_value: "102380183019283012830128309128301983")
+        expect(transfer.reload.store_requested).to be false
+      end
+    end
+
   end
 
   context "we are the to_node" do
--- a/spec/models/replication_transfer_spec.rb
+++ b/spec/models/replication_transfer_spec.rb
@@ -29,43 +29,6 @@ describe ReplicationTransfer, type: :model do
     expect(Fabricate(:replication_transfer, bag: bag, from_node: node, to_node: node)).to be_valid
   end
 
-  context "we are the from_node" do
-    it "add to_node to bag.replicating_nodes when stored" do
-      r = Fabricate(:replication_transfer, stored: false, from_node: Fabricate(:local_node))
-      expect(r.update(stored: true)).to be true
-      expect(r.bag.replicating_nodes).to include r.to_node
-    end
-
-    context "when fixity_value set" do
-      let(:correct_fixity) { "98734723942304820348203840238402983409283408234" }
-      let(:fixity_alg) { Fabricate(:fixity_alg) }
-      let(:bag) {
-        bag = Fabricate(:bag_without_digests)
-        bag.message_digests << Fabricate.build(:message_digest, fixity_alg: fixity_alg, value: correct_fixity)
-        bag.save!
-        bag
-      }
-      let(:transfer) { Fabricate(:replication_transfer,
-        fixity_alg: fixity_alg,
-        bag: bag,
-        from_node: Fabricate(:local_node))
-      }
-      it "sets store_requested->true when correct" do
-        transfer.update!(fixity_value: correct_fixity)
-        expect(transfer.reload.store_requested).to be true
-      end
-      it "cancels with fixity_reject when incorrect" do
-        transfer.update!(fixity_value: "123908102831028301820398120398102830129830")
-        expect(transfer.reload.cancelled).to be true
-        expect(transfer.reload.cancel_reason).to eql("fixity_reject")
-      end
-      it "does not set store_requested when incorrect" do
-        transfer.update!(fixity_value: "102380183019283012830128309128301983")
-        expect(transfer.reload.store_requested).to be false
-      end
-    end
-
-  end
 
   context "we are the to_node" do
     before(:each) do
@@ -77,12 +40,7 @@ describe ReplicationTransfer, type: :model do
       expect(r.bag_man_request).to be_valid
       expect(r.bag_man_request).to have_received(:begin!)
     end
-    it "calls bag_man_request.okay_to_preserve when store_requested" do
-      r = Fabricate(:replication_transfer, store_requested: false, to_node: Fabricate(:local_node))
-      allow(r.bag_man_request).to receive(:okay_to_preserve!)
-      r.update(store_requested: true)
-      expect(r.bag_man_request).to have_received(:okay_to_preserve!)
-    end
+
     it "cancels bag_man_request when cancelled" do
       r = Fabricate(:replication_transfer, to_node: Fabricate(:local_node))
       allow(r.bag_man_request).to receive(:cancel!)
--- a/spec/services/replication_transfer_updater_spec.rb
+++ b/spec/services/replication_transfer_updater_spec.rb
@@ -1,0 +1,395 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
+require 'rails_helper'
+
+describe ReplicationTransferUpdater do
+
+  def params(record, changes)
+    updated = ReplicationTransferAdapter.from_model(record).to_params_hash
+    updated.merge!(changes)
+  end
+
+  describe "::build_update" do
+    let(:record) { Fabricate(:replication_transfer) }
+    it "returns a subclass of ReplicationTransferUpdate" do
+      allow(ReplicationTransferUpdater::CancelUpdate).to receive(:matching_update?).and_return true
+      expect(described_class.build_update(record,{}).class).to eql(ReplicationTransferUpdater::CancelUpdate)
+    end
+    it "returns a generic ReplicationTransferUpdate when no specific matches exist" do
+      expect(described_class.build_update(record,{}).class).to eql(ReplicationTransferUpdater::ReplicationTransferUpdate)
+    end
+  end
+
+
+  describe ReplicationTransferUpdater::ReplicationTransferUpdate do
+    describe "#we_requested?" do
+      let(:our_transfer) { Fabricate.build(:replication_transfer, from_node: Fabricate.build(:local_node)) }
+      let(:their_transfer) { Fabricate.build(:replication_transfer, from_node: Fabricate.build(:node)) }
+      it "is true if record.from_node is the local node" do
+        expect(described_class.new(our_transfer, nil).we_requested?).to be true
+      end
+      it "is false if it was created by another" do
+        expect(described_class.new(their_transfer, nil).we_requested?).to be false
+      end
+    end
+    describe "#we_replicating?" do
+      let(:is_to_node_record) { Fabricate.build(:replication_transfer, to_node: Fabricate.build(:local_node)) }
+      let(:is_not_to_node_record) { Fabricate.build(:replication_transfer, to_node: Fabricate.build(:node)) }
+      it "is true if record.to_node is the local node" do
+        expect(described_class.new(is_to_node_record, nil).we_replicating?).to be true
+      end
+      it "is false if record.to_node is another node" do
+        expect(described_class.new(is_not_to_node_record, nil).we_replicating?).to be false
+      end
+    end
+  end
+
+
+  describe ReplicationTransferUpdater::CancelUpdate do
+    describe "::matching_update?" do
+      let(:record) { Fabricate(:replication_transfer, cancelled: false, cancel_reason: nil) }
+      let(:changes) { {cancelled: true, cancel_reason: "test"} }
+      it "matches old.cancelled == false, new.cancelled == true" do
+        expect(described_class.matching_update?(record, params(record, changes))).to be true
+      end
+    end
+
+    # Expects let(:record) and let(:changes)
+    shared_examples "CancelUpdate#update" do
+      let(:update_poro) { described_class.new(record, params(record, changes))  }
+      it "cancels the record" do
+        update_poro.update
+        expect(record.cancelled).to be true
+      end
+      it "sets the cancel reason" do
+        update_poro.update
+        expect(record.cancel_reason).to eql("test")
+      end
+      it "returns true on success" do
+        allow(record).to receive(:cancel!).and_return true
+        expect(update_poro.update).to be true
+      end
+      it "returns false on failure" do
+        allow(record).to receive(:cancel!).and_return false
+        expect(update_poro.update).to be false
+      end
+    end
+
+    describe "#update" do
+      let(:changes) { {cancelled: true, cancel_reason: "test"} }
+      context "local->other" do
+        let(:record) {
+          Fabricate(:replication_transfer,
+            from_node: Fabricate(:local_node),
+            to_node: Fabricate(:node),
+            cancelled: false,
+            cancel_reason: nil)}
+        it_behaves_like "CancelUpdate#update"
+      end
+      context "local->local" do
+        let(:record) {
+          Fabricate(:replication_transfer,
+            from_node: Fabricate(:local_node),
+            to_node: Node.local_node!,
+            cancelled: false,
+            cancel_reason: nil)}
+        it_behaves_like "CancelUpdate#update"
+      end
+      context "other->other" do
+        let(:record) {
+          Fabricate(:replication_transfer,
+            from_node: Fabricate(:node),
+            to_node: Fabricate(:node),
+            cancelled: false,
+            cancel_reason: nil)}
+        it_behaves_like "CancelUpdate#update"
+      end
+      context "other->local" do
+        let(:record) {
+          Fabricate(:replication_transfer,
+            from_node: Fabricate(:node),
+            to_node: Fabricate(:local_node),
+            cancelled: false,
+            cancel_reason: nil)}
+        it_behaves_like "CancelUpdate#update"
+      end
+    end
+  end
+
+
+  describe ReplicationTransferUpdater::FixityValueUpdate do
+    let(:fixity_alg) { Fabricate(:fixity_alg) }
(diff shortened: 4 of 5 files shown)


malakai97:
Historical note: the first build of this failed, see #70.

app/models/replication_transfer.rb
```diff
@@ -25,7 +25,7 @@ def cancel!(reason)
     unless cancelled
       transaction do
         update!(cancelled: true, cancel_reason: reason)
-        bag_man_request.cancel!(reason)
+        bag_man_request&.cancel!(reason)
```
darrenleeweber: I'm not familiar with this syntax - is this a typo?  (If it's a typo, the specs need to be improved to ensure this is covered.)

app/models/replication_transfer.rb
```diff
@@ -99,6 +100,25 @@ def cancel!(reason)
 
   private
 
+  # True if we created this request
+  def we_requested
+    from_node && from_node.namespace == Rails.configuration.local_namespace
+  end
```
dazza-codes: Consider the idiom that boolean methods end with `?`.  Possible name alternatives for this method might be `we_requested?` or `our_transfer?` or `my_transfer?` or `we_initiated?` or `we_own_transfer?` etc.

app/models/replication_transfer.rb
darrenleeweber: Never mind, now I see it's new to ruby 2.3.x, i.e.
http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/

spec/models/replication_transfer_spec.rb
```diff
@@ -35,6 +35,36 @@
...
+
+    context "when fixity_value set" do
+      let(:correct_fixity) { "98734723942304820348203840238402983409283408234" }
+      let(:fixity_alg) { Fabricate(:fixity_alg) }
+      let(:bag) {
+        bag = Fabricate(:bag_without_digests)
+        bag.message_digests << Fabricate.build(:message_digest, fixity_alg: fixity_alg, value: correct_fixity)
+        bag.save!
+        bag
+      }
```
dazza-codes: On a multi-line `let`, use the `begin ... end` block idiom rather than `{ ... }`, e.g.

``` ruby
let(:bag) begin
  # ...
end
```

spec/models/replication_transfer_spec.rb
```diff
@@ -35,6 +35,36 @@
...
+        expect(transfer.reload.cancelled).to be true
+        expect(transfer.reload.cancel_reason).to eql("fixity_reject")
+      end
+      it "does not set store_requested when incorrect" do
+        transfer.update!(fixity_value: "102380183019283012830128309128301983")
+        expect(transfer.reload.store_requested).to be false
+      end
+    end
+
   end
```
dazza-codes: Do some other specs ensure all the paths are covered?  I'm not clear about what the code coverage is here and whether all the permutations of the conditional expressions are exercised.

spec/models/replication_transfer_spec.rb
```diff
@@ -35,6 +35,36 @@
...
+        from_node: Fabricate(:local_node))
+      }
+      it "sets store_requested->true when correct" do
+        transfer.update!(fixity_value: correct_fixity)
+        expect(transfer.reload.store_requested).to be true
+      end
+      it "cancels with fixity_reject when incorrect" do
+        transfer.update!(fixity_value: "123908102831028301820398120398102830129830")
+        expect(transfer.reload.cancelled).to be true
+        expect(transfer.reload.cancel_reason).to eql("fixity_reject")
```
dazza-codes: Do you want `to eql` or `to eq` here?

malakai97: We don't want the type conversion.

dazza-codes: You've lost me, what type conversion is involved?

spec/models/replication_transfer_spec.rb
malakai97: The short answer is probably not.  
The longer answer is I agree that this isn't clear and isn't that easy to follow.  This and the other hooks should probably be redesigned into a poro.  The controller should send the update hash to that poro, and then the poro will perform the correct kind of update.  [this comment copied below]

spec/models/replication_transfer_spec.rb
malakai97: eq converts types.  eql does not.

malakai97:
This and the other hooks should probably be redesigned into a poro.  The controller should send the update hash to that poro, and then the poro will perform the correct kind of update. 

Stay tuned.

dazza-codes:
Tried to enable a code climate review of this PR to get a report on code coverage, but it's not working as expected.

dazza-codes:
Perhaps the code climate integration will only work on PRs for the master branch.  I'm working through running it on my laptop.  I really want to see the code test coverage.

I'm using the CLI i.e. https://github.com/codeclimate/codeclimate

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+  def self.update(record, params)
+    run_update(build_update(record,params))
+  end
+
+  def self.build_update(record, params)
+    [CancelUpdate, FixityValueUpdate, StoreRequestedUpdate, StoredUpdate].each do |update_class|
+      return update_class.new(record, params) if update_class.matching_update?(record, params)
+    end
+    return ReplicationTransferUpdate.new(record, params)
+  end
```
dazza-codes: build_update looks like an obscure way to create a `case` statement that returns one update handler.  If so, it would be much clearer as a `case` statement.  If not, please add comments to explain what it does.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  def self.run_update(update)
+    update.update
+  end
```
dazza-codes, line 28: `run_update` is superfluous as this one additional statement would be clearer if it were added to the end of the `update` method.  Unless `run_update` is actually called from anywhere else other than `update`?  That is, what about:

``` ruby
def self.update(record, params)
  build_update(record,params).update
end
```

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  class CancelUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.respond_to?(:cancelled) &&
+        record.cancelled == false &&
+        params[:cancelled] == true
+    end
+    def update
+      @record.cancel!(@params[:cancel_reason])
+    end
```
dazza-codes, line 62: The expected behavior of all the `update` methods is not entirely clear.  The parent class is using only `update` and not `update!`, so it will not raise an exception.  However, this override method is using `cancel!`, so it could raise an exception.  In the class below, it seems to return an update result, so there is some idea of how it should behave.  The behavior of the method through each of the classes that implement it should conform to a consistent, clearly defined role behavior.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+
+  class FixityValueUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.respond_to?(:fixity_value) &&
+        record.fixity_value.nil? &&
+        params[:fixity_value] != nil
+    end
```
dazza-codes, line 70: Maybe `!params[:fixity_value].blank?` covers all the bases?  (It may depend on what any adapters do to the params.)

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def update
+      result = @record.update(@params)
+      if result && we_requested?
+        request_storage
+      end
+      return result
+    end
```
dazza-codes, line 78: In this update, I'm a little confused about when and how many times it is updating the record and why it updates the record each time.  It's possible that if the params include a fixity value, it could verify it first before doing any updates at all?  If that fails, the only update required is to cancel it?

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+      if @params[:fixity_value] == get_digest.value
+        @record.update(store_requested: true)
+      else
+        @record.cancel!("fixity_reject")
+      end
+    end
+
+    def get_digest
+      @record.bag.message_digests.where(fixity_alg_id: @record.fixity_alg_id).first
+    end
```
dazza-codes, line 92: It seems that the `get_digest` method could include the `.value` call to simply return the fixity hash.  It would help if all these methods contained YARD documentation about `@return` types.  In this case, it seems to return a record and then `request_storage` has to use the `.value` accessor to get at the fixity value.  Moreover, that's all it ever does with the fixity record, so really what this class requires is a method to get that value, not the entire record.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+  end
+
+
+  class StoreRequestedUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.respond_to?(:store_requested) &&
+        record.store_requested == false &&
+        params[:store_requested] == true
```
dazza-codes: Does an adapter ensure that `params[:store_requested]` will be a boolean value?  Often an HTTP param is a string, like `'true'`.  If it's a string, this class would never match any update request.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+        preserve_bag
+      end
+      return result
+    end
+
+    private
+
+    def preserve_bag
+      @record.bag_man_request&.okay_to_preserve!
+    end
```
dazza-codes, line 114: This update is a little obscure to a new reader of this code.  We have to assume that the `.okay_to_preserve!` method will do the right thing.  The name of the method makes it sound more like a query method (which would be `.okay_to_preserve?`) than a command method (which might be `preserve!`).

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  end
+
+
+  class StoredUpdate < ReplicationTransferUpdate
+    def self.matching_update?(record, params)
+      record.respond_to?(:stored) &&
+        record.stored == false &&
+        params[:stored] == true
+    end
```
dazza-codes, line 123: Again, assuming `params[:stored]` will be a boolean instead of a string value.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,135 @@
...
+
+  end
+
+
+
+
+
+
+end
```
dazza-codes: Lots of superfluous white space at the end of this class.

dazza-codes:
Since we can't run code-climate service on this github PR, I've run `reek` on some files involved in this PR and here are the results:

```
$ reek app/controllers/replication_transfers_controller.rb 
app/controllers/replication_transfers_controller.rb -- 5 warnings:
  [7]:InstanceVariableAssumption: ReplicationTransfersController assumes too much for instance variable @page [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [7]:InstanceVariableAssumption: ReplicationTransfersController assumes too much for instance variable @page_size [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [7]:InstanceVariableAssumption: ReplicationTransfersController assumes too much for instance variable @replication_transfer [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [7]:InstanceVariableAssumption: ReplicationTransfersController assumes too much for instance variable @requester [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [80]:UtilityFunction: ReplicationTransfersController#create_params doesn't depend on instance state (maybe move it to another class?) [https://github.com/troessner/reek/blob/master/docs/Utility-Function.md]

$ reek app/models/replication_transfer.rb 
app/models/replication_transfer.rb -- 7 warnings:
  [120]:NilCheck: ReplicationTransfer#add_request_if_needed performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
  [28]:NilCheck: ReplicationTransfer#cancel! performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
  [7]:PrimaDonnaMethod: ReplicationTransfer has prima donna method `cancel!` [https://github.com/troessner/reek/blob/master/docs/Prima-Donna-Method.md]
  [89, 90, 91]:RepeatedConditional: ReplicationTransfer tests id.blank? at least 3 times [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
  [92, 93, 94]:RepeatedConditional: ReplicationTransfer tests v.blank? at least 3 times [https://github.com/troessner/reek/blob/master/docs/Repeated-Conditional.md]
  [78, 79, 80, 81, 82]:UncommunicativeVariableName: ReplicationTransfer has the variable name 'r' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md]
  [92, 93, 94]:UncommunicativeVariableName: ReplicationTransfer has the variable name 'v' [https://github.com/troessner/reek/blob/master/docs/Uncommunicative-Variable-Name.md]

$ reek app/services/
app/services/replication_transfer_updater.rb -- 21 warnings:
  [40]:InstanceVariableAssumption: ReplicationTransferUpdater::CancelUpdate assumes too much for instance variable @params [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [40]:InstanceVariableAssumption: ReplicationTransferUpdater::CancelUpdate assumes too much for instance variable @record [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [52]:InstanceVariableAssumption: ReplicationTransferUpdater::FixityValueUpdate assumes too much for instance variable @params [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [52]:InstanceVariableAssumption: ReplicationTransferUpdater::FixityValueUpdate assumes too much for instance variable @record [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [83]:InstanceVariableAssumption: ReplicationTransferUpdater::StoreRequestedUpdate assumes too much for instance variable @params [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [83]:InstanceVariableAssumption: ReplicationTransferUpdater::StoreRequestedUpdate assumes too much for instance variable @record [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [107]:InstanceVariableAssumption: ReplicationTransferUpdater::StoredUpdate assumes too much for instance variable @params [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [107]:InstanceVariableAssumption: ReplicationTransferUpdater::StoredUpdate assumes too much for instance variable @record [https://github.com/troessner/reek/blob/master/docs/Instance-Variable-Assumption.md]
  [40]:IrresponsibleModule: ReplicationTransferUpdater::CancelUpdate has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [52]:IrresponsibleModule: ReplicationTransferUpdater::FixityValueUpdate has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [23]:IrresponsibleModule: ReplicationTransferUpdater::ReplicationTransferUpdate has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [83]:IrresponsibleModule: ReplicationTransferUpdater::StoreRequestedUpdate has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [107]:IrresponsibleModule: ReplicationTransferUpdater::StoredUpdate has no descriptive comment [https://github.com/troessner/reek/blob/master/docs/Irresponsible-Module.md]
  [42]:ManualDispatch: ReplicationTransferUpdater::CancelUpdate#self.matching_update? manually dispatches method call [https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md]
  [54]:ManualDispatch: ReplicationTransferUpdater::FixityValueUpdate#self.matching_update? manually dispatches method call [https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md]
  [85]:ManualDispatch: ReplicationTransferUpdater::StoreRequestedUpdate#self.matching_update? manually dispatches method call [https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md]
  [109]:ManualDispatch: ReplicationTransferUpdater::StoredUpdate#self.matching_update? manually dispatches method call [https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md]
  [55]:NilCheck: ReplicationTransferUpdater::FixityValueUpdate#self.matching_update? performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
  [35]:NilCheck: ReplicationTransferUpdater::ReplicationTransferUpdate#we_replicating? performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
  [32]:NilCheck: ReplicationTransferUpdater::ReplicationTransferUpdate#we_requested? performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
  [101]:NilCheck: ReplicationTransferUpdater::StoreRequestedUpdate#preserve_bag performs a nil-check [https://github.com/troessner/reek/blob/master/docs/Nil-Check.md]
```

Consider these and see if they can help to improve the code.  It's not a requirement that the code should pass everything identified by reek, just consider it useful feedback that might prompt some better solutions.  If no better alternative solution presents itself and all the warnings are considered, that's good enough.

dazza-codes:
Consider a decorator pattern to handle some of the delegation for `@record` methods, see
http://www.thegreatcodeadventure.com/rails-refactoring-part-iii-the-decorator-pattern/

It might look something like:

``` ruby
class ReplicationTransferUpdater < SimpleDelegator
  # etc. but now the @record is the implicit object for methods
end

# elsewhere
ReplicationTransferUpdater.new(record).update
```

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,147 @@
...
+    def update
+      @record.update(@params)
+    end
+    def we_requested?
+      @record.from_node&.namespace == Rails.configuration.local_namespace
+    end
+    def we_replicating?
+      @record.to_node&.namespace == Rails.configuration.local_namespace
+    end
+  end
```
dazza-codes, line 52: Add `attr_reader :record, params` so that this and all subclasses can drop the `@` for these instance vars.

app/services/replication_transfer_updater.rb
```diff
@@ -0,0 +1,150 @@
...
+    def we_requested?
+      record.from_node&.namespace == Rails.configuration.local_namespace
+    end
+    def we_replicating?
+      record.to_node&.namespace == Rails.configuration.local_namespace
+    end
+
+    private
+    attr_reader :record, :params
+  end
```
dazza-codes, line 52: Is the `Rails.configuration.local_namespace` always going to be lower case because it is coerced when it is initialized or do all references to it when compared to a `Node.namespace` need to be `downcase`?

dazza-codes: Options might be:

``` ruby
record.to_node&.namespace == Rails.configuration.local_namespace.downcase
# or
record.to_node&.namespace =~ /^#{Rails.configuration.local_namespace}$/i
# or use the ruby start/end string regex chars instead of ^$
```

app/services/replication_transfer_updater.rb
malakai97: One day we'll all look back on this one and laugh.

app/services/replication_transfer_updater.rb
malakai97: Classes using Node shouldn't have to know if this is the case or not.  I'd waffled between comparing the namespaces as is in the code now, and comparing the node identity, i.e.

``` ruby
record.to_node == Node.local_node!
```

Where ::local_node is:

``` ruby
  def self.local_node!
    self.find_by_namespace!(Rails.configuration.local_namespace)
  end
```

One of the issues with this method is it is always going to make a database call.  Perhaps an instance method `Node#is_local_node?` would be a better idiom.  Then any downcasing logic would go there, where it belongs.  Thoughts?

dazza-codes: Yea, consolidate the string case management as much as possible.  It's too easy to forget about it and run into obscure bugs just because the case doesn't match up.

dazza-codes:
Considering

``` sh
$ reek app/services/
app/services/replication_transfer_updater.rb -- 8 warnings:
  [58]:ManualDispatch: ReplicationTransferUpdater::CancelUpdate#self.matching_update? manually dispatches method call [https://github.com/troessner/reek/blob/master/docs/Manual-Dispatch.md]
# ... snipped ...
```

It's complaining about code patterns like this one:

``` ruby
record.respond_to?(:stored) && record.stored == false && param[:stored] == true
```

So, wondering whether all the `.respond_to?` methods are a bit too paranoid?  If the db-schema defines these fields, the record should respond accordingly.  So, knowing the schema should suffice and we can throw away all the double checking and just call the method.  (Of course, assuming that `record` is an `ActiveRecord` instance.)

Another idea is something like:

``` ruby
record.stored != param[:stored]
# or go one step further
record.stored != param[:stored] && param[:stored] == true
```

(Some rspec unit tests could confirm that these should be booleans.)

The reek complaints disappear for:

``` ruby
  # An update in which the request is cancelling the record
  class CancelUpdate < ReplicationTransferUpdate
    def self.matching_update?(record, params)
      cancelled = params[:cancelled]
      record.cancelled != cancelled && cancelled == true
    end
    def update
      record.cancel!(params[:cancel_reason])
    end
  end
```

Similar changes:

``` ruby
      record.fixity_value.nil? && params[:fixity_value]
# ...
      store_requested = params[:store_requested]
      record.store_requested != store_requested && store_requested == true
# ...
      stored = params[:stored]
      record.stored != stored && stored == true
# ...
```

dazza-codes:
If the `.respond_to?` methods are dropped, the rspec needs to be modified, like:

```
  describe "::build_update" do
    let(:transfer) { Fabricate.build(:replication_transfer) }
    it "returns a subclass of ReplicationTransferUpdate" do
      params = {cancelled: true}
      allow(ReplicationTransferUpdater::CancelUpdate).to receive(:matching_update?).and_return true
      expect(described_class.build_update(transfer, params).class).to eql(ReplicationTransferUpdater::CancelUpdate)
    end
    it "returns a generic ReplicationTransferUpdate when no specific matches exist" do
      params = {}
      expect(described_class.build_update(transfer, params).class).to eql(ReplicationTransferUpdater::ReplicationTransferUpdate)
    end
  end
```

malakai97:
I still disagree that the manual dispatch smell applies here, but removing them because they're paranoid is a clear case.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+        expect(record.cancel_reason).to eql("test")
+      end
+      it "returns true on success" do
+        allow(record).to receive(:cancel!).and_return true
+        expect(update.update).to be true
+      end
+      it "returns false on failure" do
+        allow(record).to receive(:cancel!).and_return false
+        expect(update.update).to be false
+      end
```
dazza-codes, line 78: Lines 59-66 can be a global shared_example.

Do these mocks like `allow(record).to receive(:cancel!).and_return false` represent the actual behavior of the method?  Often an active record action with a `!` will raise an exception on failure, e.g.
http://apidock.com/rails/ActiveRecord/Base/save!
I think this is mocking https://github.com/dpn-admin/dpn-server/blob/feature/cancel_reason_detail/app/models/replication_transfer.rb#L24-L32

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+  }
+  let(:transfer) { Fabricate(:replication_transfer,
+    fixity_alg: fixity_alg,
+    bag: bag,
+    from_node: Fabricate(:local_node))
+  }
+
+  describe ReplicationTransferUpdater::FixityValueUpdate do
+    let(:update) { described_class.new(record, updated) }
+    let(:fixity_alg) { Fabricate(:fixity_alg) }
```
dazza-codes, line 124: This `fixity_alg` duplicates the same thing in the scope above, so it should be OK to remove it in this scope.

app/services/replication_transfer_updater.rb
malakai97: Moved this into its own issue.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+  end
+
+
+  let(:fixity_alg) { Fabricate(:fixity_alg) }
+  let(:bag) {
+    bag = Fabricate(:bag_without_digests)
+    bag.message_digests << Fabricate.build(:message_digest, fixity_alg: fixity_alg, value: correct_fixity)
+    bag.save!
+    bag
+  }
```
dazza-codes: Can't find a corresponding 'correct_fixity' value.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+      it "returns false on failure" do
+        allow(record).to receive(:cancel!).and_return false
+        expect(update.update).to be false
+      end
+    end
+  end
+
+
+  describe ReplicationTransferUpdater::FixityValueUpdate do
+    let(:update) { described_class.new(record, updated) }
```
dazza-codes: It might help any new reader of this code (it will be read plenty), to use `updater` and then see like `updater.update` instead of `update.update`.  Also, wondering whether the common `update.update` calls might also assert that the updates are a success, like `expect(updater.update).to be true`?

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+        end
+      end
+      context "we did not request this transfer" do
+        let(:record) { Fabricate(:replication_transfer, bag: bag, fixity_alg: fixity_alg,
+          fixity_value: nil)}
+        it_behaves_like "fixity_value_update_basics"
+        it "does not cancel the record" do
+          update.update
+          expect(record.cancelled).to be false
+        end
```
dazza-codes, line 162: Thinking out loud - so this spec is say checking what happens when a replicating node is doing an update for some record data that was retrieved from the admin node.  Let's say the admin node has already set the cancelled to true.  In this case, the replicating node should do the update and then we can `expect(record.cancelled).to be true`.  We might need to add this spec?

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+        it "does not set store_requested" do
+          update.update
+          expect(record.store_requested).to be false
+        end
```
dazza-codes, line 166: Likewise - we might need a spec to test that a replicating node can update `store_requested` to true when it is just saving data retrieved from another admin node that has already set this value.

malakai97: If I'm understanding you correctly this one is covered in `"we are not replicating"`

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+      end
+      context "we requested, fixity correct" do
+        let(:record) { Fabricate(:replication_transfer, bag: bag, fixity_alg: fixity_alg,
+          fixity_value: nil, from_node: Fabricate(:local_node))}
+        let(:changes) { { fixity_value: fixity_value} }
+        it_behaves_like "fixity_value_update_basics"
+        it "sets store_requested->true" do
+          update.update
+          expect(record.store_requested).to be true
+        end
```
dazza-codes, line 173: In this spec ^^, can we add some `expect` statements that check one or more methods that validate the fixity value from the params against a known good fixity value?

spec/services/replication_transfer_updater_spec.rb
malakai97: This ends up being a check that a field doesn't change.  Pulling that thread leads to checking that every field we don't care about doesn't change.  That's not completely unreasonable to have, and could be accomplished by comparing against an expected record.  

However, I think that ends up duplicating the tests that are on the ReplicationTransfer model itself, as that's where the validation for this kind of stuff lives.  I think our tests become more fragile that way, too.  DIdn't seem like a tradeoff we wanted.  Do you disagree?

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+      context "we requested, fixity incorrect" do
+        let(:record) { Fabricate(:replication_transfer, bag: bag, fixity_alg: fixity_alg,
+          fixity_value: nil, from_node: Fabricate(:local_node))}
+        let(:changes) { { fixity_value: "93847238479237490238"} }
+        it_behaves_like "fixity_value_update_basics"
+        it "cancels with fixity_reject" do
+          update.update
+          expect(record.cancelled).to be true
+          expect(record.cancel_reason).to eql("fixity_reject")
+        end
```
dazza-codes, line 185: In this spec ^^, can we add some expect statements that check one or more methods that validate the fixity value from the params against a known good fixity value?  The expect should check that the fixity validation failed.

spec/services/replication_transfer_updater_spec.rb
malakai97: See  `it_behaves_like "fixity_value_update_basics"` within the describe block, line 121

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+          expect(record.cancelled).to be false
+        end
+        it "does not set stored" do
+          update.update
+          expect(record.stored).to be false
+        end
+      end
+      context "we are not replicating" do
+        let(:record) { Fabricate(:replication_transfer, store_requested: false, to_node: Fabricate(:node)) }
+        it_behaves_like "store_requested_update"
```
dazza-codes: below, typo in `presevation` -> `preservation`

spec/services/replication_transfer_updater_spec.rb
malakai97: Similarly, see line 135

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+          bag_man_request = double(:bag_man_request)
+          allow(record).to receive(:bag_man_request).and_return(bag_man_request)
+          expect(bag_man_request).to_not receive(:okay_to_preserve!)
+          update.update
+        end
+      end
+      context "we are replicating" do
+        let(:record) { Fabricate(:replication_transfer, store_requested: false, to_node: Fabricate(:local_node)) }
+        it_behaves_like "store_requested_update"
+        it "requests presevation" do
```
dazza-codes: `preservation`

spec/services/replication_transfer_updater_spec.rb
dazza-codes: I'm struggling to see the fixture data definitions for the to_node and from_node and whether or not the setup for a given spec context has the local node equal to one of these.  So, for any context like this one, `"we did not request this transfer"`, I guess I want to see a spec that tests something like `local_node != from_node`, to confirm that the specs are setup correctly?

spec/services/replication_transfer_updater_spec.rb
dazza-codes: But that stuff doesn't check any methods that do the fixity value comparisons, it just checks for db changes.  I'm looking for more detailed specs on the actual fixity equality/comparison method.

dazza-codes: Somewhere, the specs should confirm that the fixtures are setup correctly by ensuring that 
`params[:fixity_value] == correct_fixity` will be true (or false, as required) in `FixityValueUpdate#request_storage`.

dazza-codes: When I look at line 119 and see `fixity_value: nil` and then at line 120 and see `fixity_value: fixity_value`, I think, shoot, they are not equal.  Then I wonder where the `correct_fixity` is coming from, you know?

spec/services/replication_transfer_updater_spec.rb
dazza-codes: `fixity_value` in line 133 is `nil` and line 134 is a string/integer.  They seem to be different, but there's nothing in the specs to confirm that FixityValueUpdate will determine that they are in fact different.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+        it_behaves_like "store_requested_update"
+        it "does not request preservation" do
+          bag_man_request = double(:bag_man_request)
+          allow(record).to receive(:bag_man_request).and_return(bag_man_request)
+          expect(bag_man_request).to_not receive(:okay_to_preserve!)
+          update.update
+        end
+      end
+      context "we are replicating" do
+        let(:record) { Fabricate(:replication_transfer, store_requested: false, to_node: Fabricate(:local_node)) }
```
dazza-codes: ^^ Here I can clearly see the `to_node == local_node` idea in this fixture and it corresponds with the label for the context.  That is, a replication node is a local_node == to_node.  (Did I miss this stuff when I looked at fixtures above?  Or did github already update your changes while I'm reviewing all this stuff?)

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+    let(:changes) { {stored: true} }
+    describe "::matching_update?" do
+      let(:record) { Fabricate(:replication_transfer, stored: false) }
+      it "matches old.stored == false, new.stored == true" do
+        expect(described_class.matching_update?(record, updated)).to be true
+      end
+    end
+    describe "#update" do
+      context "we did not request this transfer" do
+        let(:record) { Fabricate(:replication_transfer, stored: false) }
```
dazza-codes: Could this be explicit about setting the `from_node` to something other than `local_node` and perhaps also setting the `to_node` to something else (which might be local_node or another node altogether).  This prompts me to think there are actually 3 contexts:
- `local_node` == `from_node` (!= `to_node`) - so all the admin node operations apply
- `local_node` == `to_node` (!= `from_node`) - so all the replicating node operations apply
- `local_node` is not either `to_node` or `from_node` - this should do only a generic update

dazza-codes: If this is not explicit here in this context, we have to go look it up in the fixture code somewhere else.  When it's somewhere else, anyone who changes that code may not be aware of the impact of that change on this particular context.

dazza-codes: It's dawning on me that there could be a few global `let` statements, maybe:

``` ruby
let(:admin_record) { Fabricate(:replication_transfer, from_node: Fabricate(:local_node), to_node: 'nodeX') }
let(:slave_record) { Fabricate(:replication_transfer, from_node: 'nodeX', to_node: Fabricate(:local_node)) }
let(:anon_record) { Fabricate(:replication_transfer, from_node: 'nodeX', to_node: 'nodeY') }
```

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+          expect(update.update).to be true
+        end
+        it "returns false on failure" do
+          allow(record).to receive(:update).and_return false
+          expect(update.update).to be false
+        end
+        it "does not cancel the record" do
+          update.update
+          expect(record.cancelled).to be false
+        end
```
dazza-codes: Lines 245-256 can be a global shared_example.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
```
dazza-codes, line 354: Lines 222-233 can be a global shared_example.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
```
dazza-codes, line 278: Lines 165-176 can be a global shared_example.

spec/services/replication_transfer_updater_spec.rb
```diff
@@ -0,0 +1,266 @@
...
+          expect(record.fixity_value).to eql(changes[:fixity_value])
+        end
+        it "returns true on success" do
+          allow(record).to receive(:update).and_return true
+          expect(update.update).to be true
+        end
+        it "returns false on failure" do
+          allow(record).to receive(:update).and_return false
+          expect(update.update).to be false
+        end
```
dazza-codes, line 155: Lines 95-102 can be a global shared_example.

PR #76, opened 2016-09-21
Display resque schedule, do not load unused redis

malakai97:
Correctly display resque schedule. 
Do not attempt to load redis,resque if not in use as defined by config/dpn.yml


Pcolar:
Are there any cases where we would selectively schedule these events for some nodes?
If not a current use case, consider as a future enhancement.

PR #77, opened 2016-09-21
Feature/cancel reason detail

malakai97:
Branched off of #69, so not ready to merge until that is merged.

### DIFF
--- a/app/adapters/restore_transfer_adapter.rb
+++ b/app/adapters/restore_transfer_adapter.rb
@@ -13,6 +13,7 @@ class RestoreTransferAdapter < ::AbstractAdapter
   map_bool    :finished, :finished
   map_bool    :cancelled, :cancelled
   map_simple  :cancel_reason, :cancel_reason
+  map_simple  :cancel_reason_detail, :cancel_reason_detail
   map_simple  :link, :link
 
   map_belongs_to :bag, :bag, sub_method: :uuid
--- a/app/models/bag_man_request.rb
+++ b/app/models/bag_man_request.rb
@@ -39,7 +39,7 @@ class BagManRequest < ActiveRecord::Base
     unless cancelled
       transaction do
         update!(cancelled: true, cancel_reason: reason)
-        replication_transfer.cancel!(reason)
+        replication_transfer.cancel!(reason, last_error)
         send_cancel(replication_transfer)
       end
     end
--- a/app/models/replication_transfer.rb
+++ b/app/models/replication_transfer.rb
@@ -21,10 +21,10 @@ class ReplicationTransfer < ActiveRecord::Base
 
 
 
-  def cancel!(reason)
+  def cancel!(reason, detail)
     unless cancelled
       transaction do
-        update!(cancelled: true, cancel_reason: reason)
+        update!(cancelled: true, cancel_reason: reason, cancel_reason_detail: detail)
         bag_man_request&.cancel!(reason)
       end
     end
@@ -80,6 +80,7 @@ class ReplicationTransfer < ActiveRecord::Base
   validates :stored,          read_only: true, on: :update, unless: proc {|r| r.stored_changed?(from: false, to: true)}
   validates :cancelled,       read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
   validates :cancel_reason,   read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
+  validates :cancel_reason_detail, read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
 
 
 
--- a/app/models/restore_transfer.rb
+++ b/app/models/restore_transfer.rb
@@ -19,9 +19,9 @@ class RestoreTransfer < ActiveRecord::Base
     Set.new [:restore_id]
   end
 
-  def cancel!(reason)
+  def cancel!(reason, detail)
     unless cancelled
-      update!(cancelled: true, cancel_reason: reason)
+      update!(cancelled: true, cancel_reason: reason, cancel_reason_detail: detail)
     end
   end
 
@@ -63,6 +63,7 @@ class RestoreTransfer < ActiveRecord::Base
   validates :finished,        read_only: true, on: :update, unless: proc {|r| r.finished_changed?(from: false, to: true)}
   validates :cancelled,       read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
   validates :cancel_reason,   read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
+  validates :cancel_reason_detail, read_only: true, on: :update, unless: proc {|r| r.cancelled_changed?(from: false, to: true)}
 
 
   ### Scopes
--- a/app/services/replication_transfer_updater.rb
+++ b/app/services/replication_transfer_updater.rb
@@ -58,7 +58,7 @@ class ReplicationTransferUpdater
       record.cancelled == false && params[:cancelled] == true
     end
     def update
-      record.cancel!(params[:cancel_reason])
+      record.cancel!(params[:cancel_reason], params[:cancel_reason_detail])
     end
   end
 
@@ -83,12 +83,13 @@ class ReplicationTransferUpdater
       if params[:fixity_value] == correct_fixity
         record.update(store_requested: true)
       else
-        record.cancel!("fixity_reject")
+        detail = "expected: '#{correct_fixity}', got: '#{params[:fixity_value]}'"
+        record.cancel!("fixity_reject", detail)
       end
     end
 
     def correct_fixity
-      record.bag.message_digests.where(fixity_alg_id: record.fixity_alg_id).first.value
+      @correct_fixity ||= record.bag.message_digests.where(fixity_alg_id: record.fixity_alg_id).first.value
     end
   end
 
--- a/db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
+++ b/db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
@@ -1,0 +1,5 @@
+class AddCancelReasonDetailToReplicationTransfer < ActiveRecord::Migration
+  def change
+    add_column :replication_transfers, :cancel_reason_detail, :string, limit: 255, null: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160811162500) do
+ActiveRecord::Schema.define(version: 20160921185130) do
 
   create_table "bag_man_requests", force: :cascade do |t|
     t.string   "source_location",                         null: false
@@ -152,21 +152,22 @@ ActiveRecord::Schema.define(version: 20160811162500) do
   end
 
   create_table "replication_transfers", force: :cascade do |t|
-    t.integer  "bag_id",                          null: false
-    t.integer  "from_node_id",                    null: false
-    t.integer  "to_node_id",                      null: false
-    t.integer  "protocol_id",                     null: false
-    t.string   "link",                            null: false
-    t.integer  "fixity_alg_id",                   null: false
+    t.integer  "bag_id",                                           null: false
+    t.integer  "from_node_id",                                     null: false
+    t.integer  "to_node_id",                                       null: false
+    t.integer  "protocol_id",                                      null: false
+    t.string   "link",                                             null: false
+    t.integer  "fixity_alg_id",                                    null: false
     t.text     "fixity_nonce"
     t.string   "fixity_value"
-    t.datetime "created_at",                      null: false
-    t.datetime "updated_at",                      null: false
-    t.string   "replication_id",                  null: false
-    t.boolean  "store_requested", default: false, null: false
-    t.boolean  "stored",          default: false, null: false
-    t.boolean  "cancelled",       default: false, null: false
+    t.datetime "created_at",                                       null: false
+    t.datetime "updated_at",                                       null: false
+    t.string   "replication_id",                                   null: false
+    t.boolean  "store_requested",                  default: false, null: false
+    t.boolean  "stored",                           default: false, null: false
+    t.boolean  "cancelled",                        default: false, null: false
     t.text     "cancel_reason"
+    t.string   "cancel_reason_detail", limit: 255
   end
 
   add_index "replication_transfers", ["replication_id"], name: "index_replication_transfers_on_replication_id", unique: true
@@ -179,18 +180,19 @@ ActiveRecord::Schema.define(version: 20160811162500) do
   end
 
   create_table "restore_transfers", force: :cascade do |t|
-    t.integer  "bag_id",                        null: false
-    t.integer  "from_node_id",                  null: false
-    t.integer  "to_node_id",                    null: false
-    t.integer  "protocol_id",                   null: false
+    t.integer  "bag_id",                                           null: false
+    t.integer  "from_node_id",                                     null: false
+    t.integer  "to_node_id",                                       null: false
+    t.integer  "protocol_id",                                      null: false
     t.string   "link"
-    t.datetime "created_at",                    null: false
-    t.datetime "updated_at",                    null: false
-    t.string   "restore_id",                    null: false
-    t.boolean  "accepted",      default: false, null: false
-    t.boolean  "finished",      default: false, null: false
-    t.boolean  "cancelled",     default: false, null: false
+    t.datetime "created_at",                                       null: false
+    t.datetime "updated_at",                                       null: false
+    t.string   "restore_id",                                       null: false
+    t.boolean  "accepted",                         default: false, null: false
+    t.boolean  "finished",                         default: false, null: false
+    t.boolean  "cancelled",                        default: false, null: false
     t.string   "cancel_reason"
+    t.string   "cancel_reason_detail", limit: 255
   end
 
   add_index "restore_transfers", ["restore_id"], name: "index_restore_transfers_on_restore_id", unique: true
--- a/spec/adapters/replication_transfer_adapter_spec.rb
+++ b/spec/adapters/replication_transfer_adapter_spec.rb
@@ -17,6 +17,7 @@ describe ReplicationTransferAdapter do
   updated_at = "2015-02-25T15:27:40Z"
   created_at = "2015-02-25T15:27:40Z"
   link = "user@herp.derp.org:/blah"
+  cancel_reason_detail = "testing is fun # comment // woo"
 
 
   before(:each) do
@@ -34,6 +35,7 @@ describe ReplicationTransferAdapter do
       store_requested: true,
       stored: false,
       cancelled: true,
+      cancel_reason_detail: cancel_reason_detail,
       cancel_reason: 'other',
       created_at: time_from_string(created_at),
       updated_at: time_from_string(updated_at)
@@ -51,6 +53,7 @@ describe ReplicationTransferAdapter do
       stored: false,
       cancelled: true,
       cancel_reason: 'other',
+      cancel_reason_detail: cancel_reason_detail,
       protocol: protocol_name,
       link: link,
       created_at: created_at,
@@ -71,6 +74,7 @@ describe ReplicationTransferAdapter do
       stored: false,
       cancelled: true,
       cancel_reason: 'other',
+      cancel_reason_detail: cancel_reason_detail,
       created_at: time_from_string(created_at),
       updated_at: time_from_string(updated_at)
     }
--- a/spec/models/bag_man_request_spec.rb
+++ b/spec/models/bag_man_request_spec.rb
@@ -59,31 +59,42 @@ describe BagManRequest, type: :model do
     end
 
     context "when not cancelled" do
+      let(:from_node_namespace) { "fake_namespace" }
+      let(:replication) { Fabricate(:replication_transfer, from_node: Fabricate(:node, namespace: from_node_namespace)) }
+      let(:request) {
+        request = Fabricate(:bag_man_request)
+        allow(request).to receive(:replication_transfer).and_return(replication)
+        request
+      }
       before(:each) do
-        @repl_mock = double(:replication_transfer)
-        allow(@repl_mock).to receive(:cancel!)
-        allow(@repl_mock).to receive(:from_node).and_return FromNode.new("fake_namespace")
-        @request = Fabricate(:bag_man_request)
-        allow(@request).to receive(:replication_transfer).and_return(@repl_mock)
         allow(Client::Repl::CancelJob).to receive(:perform_later)
       end
       it "sets the cancel_reason" do
-        @request.cancel!('testing')
-        expect(@request.reload.cancel_reason).to eql('testing')
+        expect(request.cancel_reason).not_to eql('testing')
+        expect {
+          request.cancel!('testing')
+        }.to change {
+          request.reload.cancel_reason
+        }.to eql('testing')
       end
       it "calls replication_transfer.cancel!" do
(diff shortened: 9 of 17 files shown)


dazza-codes:
Checking some files modified in this PR with `reek`; the details are snipped below, run it locally to see them.  I'll focus on the code first and look at specs later.

```
$ reek app/adapters/adapter.rb 
app/adapters/adapter.rb -- 14 warnings:
$ reek app/adapters/abstract_adapter.rb 
app/adapters/abstract_adapter.rb -- 8 warnings:
$ reek app/adapters/replication_transfer_adapter.rb 
# OK
$ reek app/adapters/restore_transfer_adapter.rb 
# OK
$ reek app/models/bag_man_request.rb 
app/models/bag_man_request.rb -- 12 warnings:
$ reek app/models/replication_transfer.rb 
app/models/replication_transfer.rb -- 7 warnings:
$ reek app/models/restore_transfer.rb 
app/models/restore_transfer.rb -- 5 warnings:
$ reek app/services/replication_transfer_updater.rb 
app/services/replication_transfer_updater.rb -- 4 warnings:
# we've probably done the best we can on this one already
$ reek db/migrate/2016092118*
db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb -- 1 warning:
db/migrate/20160921185130_add_cancel_reason_detail_to_restore_transfer.rb -- 1 warning:
```

Some of these warnings might be addressed or used to improve the code quality.  Some of them will not get resolved.

darrenleeweber:
While looking at `app/adapters/adapter.rb` and `app/adapters/abstract_adapter.rb`, it seems they could be combined into one class, unless there is a design pattern involved?  Also, it's not entirely clear why the `abstract_adapter.rb` requires both class and instance methods (it seems lines 50-57 try to create instance methods that delegate to class methods for `:simple_maps` and `:to_maps`, but they seem to be instance methods already in `adapter.rb`).

darrenleeweber:
The `AbstractAdapter#to_public_hash` seems to be memoized, but that could prevent field additions or deletions from showing up?  Perhaps it should not be memoized?

dazza-codes:
Code coverage is missing, for files touched in this PR, in:
- `AbstractAdapter#to_json`
- `AbstractAdapter#merge_strategy`
- `ReplicationTransfer#to_param`
- `RestoreTransfer#to_param`
- `ReplicationTransferUpdater#build_update`

PS, moving this concern to https://github.com/dpn-admin/dpn-server/issues/87

app/adapters/restore_transfer_adapter.rb
```diff
@@ -13,6 +13,7 @@ class RestoreTransferAdapter < ::AbstractAdapter
   map_bool    :finished, :finished
   map_bool    :cancelled, :cancelled
   map_simple  :cancel_reason, :cancel_reason
+  map_simple :cancel_reason_detail, :cancel_reason_detail
   map_simple  :link, :link
```
darrenleeweber: Adjust whitespace indentation to match.

app/models/bag_man_request.rb
```diff
@@ -39,7 +39,7 @@ def cancel!(reason)
     unless cancelled
       transaction do
         update!(cancelled: true, cancel_reason: reason)
-        replication_transfer.cancel!(reason)
+        replication_transfer.cancel!(reason, last_error)
         send_cancel(replication_transfer)
       end
     end
```
dazza-codes, line 45: Where does the `last_error` come from?  What is the relationship, if any, between the `last_error` and the `reason`?  If this is not obvious, add commentary to explain the context.

Should the `send_cancel` method check whether the background jobs are enabled?

malakai97: `last_error` is a field on the bag_man_request.  

The last_error could contain information about why the bag_man_request was cancelled; more context over always having the cancel_reason_detail be nil. The other readily-available option is to forego sharing that information with the from_node.

No, what the bag_man_request should do when cancelled isn't affected by whether the jobs are running or not.

app/models/replication_transfer.rb
```diff
@@ -21,10 +21,10 @@ def self.find_fields
...
 
-  def cancel!(reason)
+  def cancel!(reason, detail)
     unless cancelled
       transaction do
-        update!(cancelled: true, cancel_reason: reason)
+        update!(cancelled: true, cancel_reason: reason, cancel_reason_detail: detail)
         bag_man_request&.cancel!(reason)
       end
     end
```
dazza-codes, line 30: The control flow, transactionality or atomicity, of the `cancel!` operations between the `ReplicationTransfer` and the `BagManRequest` are complex because they each call the other and, without the `unless cancelled` wrapper, might incur an infinite loop.  I wonder whether the calls from one model to the other should be extracted out of the `transaction` block, to allow an atomic update of one model and then another model?  Or, perhaps this coupling between these models should be extracted from the model level up to the controller level?

app/models/restore_transfer.rb
```diff
@@ -19,9 +19,9 @@ def self.find_fields
     Set.new [:restore_id]
   end
 
-  def cancel!(reason)
+  def cancel!(reason, detail)
     unless cancelled
-      update!(cancelled: true, cancel_reason: reason)
+      update!(cancelled: true, cancel_reason: reason, cancel_reason_detail: detail)
     end
   end
```
dazza-codes, line 26: This looks clean and simple.  Just checking -- are there any associated bag transfer details to synchronize?  If there are, is this done in a controller?

app/services/replication_transfer_updater.rb
```diff
@@ -83,7 +83,7 @@ def request_storage
       if params[:fixity_value] == correct_fixity
         record.update(store_requested: true)
       else
-        record.cancel!("fixity_reject")
+        record.cancel!("fixity_reject", params[:cancel_reason_detail])
       end
```
dazza-codes, line 88: Should this method be creating the cancel reason detail, with information about why the fixity values fail to match up?

app/models/replication_transfer.rb
malakai97: I agree there must be a better strategy.  The root of the complexity seems to be the coupling between the concerns, and the fact that the concern addressed by BagManRequest are addressed via external tools at some nodes.  

Moving this into the controller level offers a fair amount of room to maneuver, but would require rewriting the Repl jobs to use the new route.  I think we'll want to look at this for a minor release soon.

app/models/restore_transfer.rb
malakai97: The simplicity is a lie! -> The automation for RestoreTransfers isn't implemented.

app/services/replication_transfer_updater.rb
malakai97: Yes!

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
```diff
@@ -0,0 +1,5 @@
+class AddCancelReasonDetailToReplicationTransfer < ActiveRecord::Migration
+  def change
+    add_column :replication_transfers, :cancel_reason_detail, :text, null: true
+  end
+end
```
dazza-codes, line 5: Is `null: true` the default?

Do we need to add `limit:` option for the length of a text field?

Same questions apply to the other migration in this PR.

db/schema.rb
```diff
@@ -99,7 +99,7 @@
     t.integer  "bag_id",        null: false
     t.integer  "fixity_alg_id", null: false
     t.text     "value",         null: false
-    t.integer  "node_id",       null: false
+    t.integer  "node_id"
     t.datetime "created_at",    null: false
   end
```
dazza-codes, line 104: Not sure why this should change here in this PR.

spec/adapters/replication_transfer_adapter_spec.rb
```diff
@@ -17,6 +17,7 @@
   updated_at = "2015-02-25T15:27:40Z"
   created_at = "2015-02-25T15:27:40Z"
   link = "user@herp.derp.org:/blah"
+  cancel_reason_detail = "testing is fun # comment // woo"
```
darrenleeweber: Could this move to a fabricator?  (Likewise for other instances in other spec files.)

app/services/replication_transfer_updater.rb
malakai97: Hmm, I don't know what information this updater would want to put into the cancel_reason_detail.  The reported and expected fixities are already available.  With future nonce support, we could report what we expected, or keep that information hidden.  Thoughts?

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
malakai97: I don't know which is the default, but I prefer to be explicit.  What do you think about the limit?  There's likely a different default for each database adapter.

spec/adapters/replication_transfer_adapter_spec.rb
malakai97: Do you mean like Fabricate(:cancel_reason_detail) ?  I'm not sure if Fabricators are built to handle plain objects, but we could certainly roll our own or just replace the string literal with an appropriate Faker call.

spec/models/bag_man_request_spec.rb
```diff
@@ -59,31 +59,42 @@
     end
 
     context "when not cancelled" do
+      let(:replication) {
+        mock = double(:replication_transfer)
+        allow(mock).to receive(:cancel!)
+        allow(mock).to receive(:from_node).and_return FromNode.new("fake_namespace")
+        mock
+      }
```
dazza-codes: ^^ use a fabricator for `replication` here?

spec/models/bag_man_request_spec.rb
```diff
@@ -59,31 +59,42 @@
...
-        @request = Fabricate(:bag_man_request)
-        allow(@request).to receive(:replication_transfer).and_return(@repl_mock)
         allow(Client::Repl::CancelJob).to receive(:perform_later)
       end
       it "sets the cancel_reason" do
-        @request.cancel!('testing')
-        expect(@request.reload.cancel_reason).to eql('testing')
+        request.cancel!('testing')
+        expect(request.reload.cancel_reason).to eql('testing')
       end
```
dazza-codes, line 79: Use a change matcher?  Something like

``` ruby
expect { request.cancel!('testing') }.to change(request.cancel_reason)
```

db/schema.rb
malakai97: This is coming from a migration in `develop`.

spec/models/bag_man_request_spec.rb
```diff
@@ -59,31 +59,42 @@
...
       it "calls replication_transfer.cancel!" do
-        expect(@repl_mock).to receive(:cancel!)
-        @request.cancel!('testing')
+        expect(replication).to receive(:cancel!)
+        request.cancel!('testing')
       end
```
dazza-codes, line 83: I'm curious about what would happen here when the `replication.cancel!` method is not stubbed out or if some specific counts were added to the specs.  I think the "circularity" issue noted above might be caught.

app/models/bag_man_request.rb
dazza-codes: I think `send_cancel` will try to add a job on the queue, but if the rescue queue is not responding, it might fail?  Is that a realistic scenario and is it handled somehow?

app/models/replication_transfer.rb
dazza-codes: OK, let's spin off an issue for the backlog, as this point is bigger than this PR alone.

app/models/restore_transfer.rb
dazza-codes: OK, this could be another issue for the backlog; it's doesn't need to be a blocker for this PR.

app/models/bag_man_request.rb
malakai97: If the jobs are disabled we use a null queue, so that'll return success.  If the jobs are enabled and resque or redis is down, that should explode which causes the transaction to be rolled back.

app/services/replication_transfer_updater.rb
dazza-codes: In slack chat we seemed to arrive at logging some details and adding a detail that simply refers to the logs (maybe add a date or other information about accessing the log).

spec/models/bag_man_request_spec.rb
malakai97: I did not know about this functionality of rspec.  However it does not seem to have been extended by rspec-rails to handle the reload of the object.

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
dazza-codes: I've bumped into some MySQL migration issues where it requires a limit on a text field.  (I haven't tried to test these migrations on MySQL yet.)  Adding an explicit limit might be good, to wrestle some control away from the specific-db defaults.

spec/adapters/replication_transfer_adapter_spec.rb
darrenleeweber: Oh, I see, this is the adapter spec, not the model.  OK.

spec/models/bag_man_request_spec.rb
```diff
@@ -59,31 +59,42 @@
...
       it "passes the reason to replication_transfer.cancel!" do
-        expect(@repl_mock).to receive(:cancel!).with('testing')
-        @request.cancel!('testing')
+        request.cancel!('testing')
+        expect(replication).to have_received(:cancel!).with('testing', anything)
+      end
```
dazza-codes, line 87: My guess is that we could add this to the top of this example (not that we should add it):

``` ruby
expect(request).to receive(:cancel!).twice.and_call_original
```

spec/models/bag_man_request_spec.rb
```diff
@@ -59,31 +59,42 @@
...
+        expect(replication).to have_received(:cancel!).with(anything, request.last_error)
       end
       it "enqueues a Client::Repl::CancelJob" do
         expect(Client::Repl::CancelJob).to receive(:perform_later)
-          .with(@repl_mock, "fake_namespace", "replicate",
+          .with(replication, "fake_namespace", "replicate",
             "update_replication", "ReplicationTransferAdapter")
-        @request.cancel!('testing')
+        request.cancel!('testing')
       end
```
dazza-codes, line 98: What happens when `Client::Repl::CancelJob#perform_later` raises an exception?  Does it cancel the transaction that cancels the replication_transfer?  If it does, should there be a `retry` block somewhere?

app/services/replication_transfer_updater.rb
malakai97: I ended up deciding to just say what we got and what we expected.  For now both of these values are in the registry anyway.  In the future, publishing the nonced fixity forces discarding the nonce and actually provides something meaningful the to_node can use out of band to try to decipher what went wrong.

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
malakai97: @Pcolar pick a number, any number...

app/models/bag_man_request.rb
dazza-codes: OK, sounds like it should be in the transaction.  Although, just adding an item to the job queue doesn't ensure that it's done.  So, wondering whether the job itself should be using the API to check or update the cancel data?  But then, if the job does issue a cancel API request, this could incur another level of loop management?

spec/models/bag_man_request_spec.rb
dazza-codes: I think it works for active record objects.

spec/models/bag_man_request_spec.rb
malakai97: The assumption is you're either cancelling the replication request because (1) the to_node cancelled it on their end and you're syncing that change, or (2) you want to cancel it for some reason.  Both of these are powered by jobs or clients, where the expectation is that a job/client will run again if it encounters an exception, up to some max number of attempts.  The code doesn't make any attempts to handle what happens after that.

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
malakai97: We haven't set limits for any of the other fields to my knowledge.

spec/models/replication_transfer_spec.rb
```diff
@@ -44,7 +44,7 @@
     it "cancels bag_man_request when cancelled" do
       r = Fabricate(:replication_transfer, to_node: Fabricate(:local_node))
       allow(r.bag_man_request).to receive(:cancel!)
-      r.cancel!('other')
+      r.cancel!('other', 'detail')
       expect(r.bag_man_request).to have_received(:cancel!)
     end
```
dazza-codes, line 49: This spec also hides the circularity in the model calls for cancel.  If the `allow` on `r.bag_man_request.cancel!` is replaced with an `expect...to receive...`, then the last line can be deleted.  It might catch if there is more than one call to `r.bag_man_request.cancel!` but I don't think it will.  There might be more than one call to `r.cancel!`, wich could be made explicit with something like:

``` ruby
expect(r.bag_man_request).to receive(:cancel!).and_call_original
expect(r).to receive(:cancel!).twice.and_call_original
r.cancel!('other', 'detail')
# remove:  expect(r.bag_man_request).to have_received(:cancel!)
```

spec/models/replication_transfer_spec.rb
```diff
@@ -140,12 +141,32 @@
...
 
 
+  describe "#cancel!" do
+    let(:transfer) { Fabricate(:replication_transfer) }
+    before(:each) { transfer.cancel!('other', 'test') }
+    it "is idempotent" do
+      expect {
+        transfer.cancel!('other', 'test')
+      }.to_not raise_error
+    end
```
dazza-codes, line 157: This example might add `expect(transfer).to receive(:cancelled).and_call_original` at the top.  If it catches on the count, could add `twice`.  Another option is to `expect(transfer.cancelled).to be true`.

dazza-codes:
In general, this change looks like it will work out OK.

I am concerned about possible circularity and tight coupling between models, so I’m suggesting ways to make this clear in the unit tests rather than masking them with stub/mock code that doesn’t fully execute the methods involved.  I’m open to spinning off issues for this, although it might be good to address it now or soon, while the issue is fresh in our minds.  Obviously, this is the first time that I’ve seen some of this code, so it’s raising some questions that are not exactly specific to this PR, but can’t be left to drift in the wind.  If there are any bugs in this coupling of the models, it would be good to identify them now.

Although I haven’t thought it all through yet, I’m leaning toward trying to decouple the ReplicationTransfer and BagManRequest (in the model code).  As I understand it, the former is truly part of the API spec, whereas the latter is part of the node-sync processes that use the API.  One way to decouple them is to move the coupling to an API controller, another way is to move all the updates for BagManRequest into the background job.  My first guess is that the latter is preferable.  In other words, it’s OK for the server to cancel a ReplicationTransfer and then, somehow, the background jobs will eventually find a way to get into a consistent state with that change (and there should be no way for the jobs to modify whether or not the transfer is canceled once the server determines that it should be canceled, so the read-only quality of the cancelled status is OK).

PS: Created https://github.com/dpn-admin/dpn-server/issues/80 to address concerns about model coupling.

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
Pcolar: We should consider the overhead of syncing a large reason field. Are we trying to encapsulate the entire failure stack or a pointer to local information?

app/models/replication_transfer.rb
dazza-codes: Extracting this issue to https://github.com/dpn-admin/dpn-server/issues/80

app/models/restore_transfer.rb
dazza-codes: Extracting this to https://github.com/dpn-admin/dpn-server/issues/81

app/services/replication_transfer_updater.rb
dazza-codes: This looks reasonable, in commit 6bad3d0

db/schema.rb
dazza-codes: This can slide, if it was just an omission from a prior commit/PR.

spec/adapters/replication_transfer_adapter_spec.rb
darrenleeweber: Deleting these comments because this is resolved without changes.

spec/models/bag_man_request_spec.rb
dazza-codes: Moving this to https://github.com/dpn-admin/dpn-server/issues/80

spec/models/replication_transfer_spec.rb
dazza-codes: Moving this to https://github.com/dpn-admin/dpn-server/issues/80

spec/models/bag_man_request_spec.rb
dazza-codes: Moving this to https://github.com/dpn-admin/dpn-server/issues/80

spec/models/bag_man_request_spec.rb
dazza-codes: Moving this to https://github.com/dpn-admin/dpn-server/issues/80

spec/models/replication_transfer_spec.rb
dazza-codes: Moving this to https://github.com/dpn-admin/dpn-server/issues/80

malakai97:
Resolved the limit issue (255, as STRING).
Squashed commits

db/migrate/20160921184937_add_cancel_reason_detail_to_replication_transfer.rb
dazza-codes: This is resolved by using a string field (255 chars).

app/models/bag_man_request.rb
dazza-codes: This can be resolved in #80

db/schema.rb
dazza-codes: This can be resolved in https://github.com/dpn-admin/dpn-server/issues/86

spec/models/bag_man_request_spec.rb
malakai97: It has no logic for getting the updated record from the database.  The rspec-rails issue is unresolved.  https://github.com/rspec/rspec-rails/issues/1173

dazza-codes: The following works, although it's a lot more verbose than what I thought would work:

``` ruby
      it "sets the cancel_reason" do
        expect(request.cancel_reason).not_to eql('testing')
        expect {
          request.cancel!('testing')
        }.to change {
          request.reload.cancel_reason
        }.to eql('testing')
      end
```

darrenleeweber: It might be shortened to:
```ruby
      it "sets the cancel_reason" do
        expect(request.cancel_reason).not_to eql('testing')
        change { request.cancel!('testing') }.to eql('testing')
      end
```

dazza-codes: I confirmed the above works as expected by changing it to `eql('what')` and that fails with

```
  1) BagManRequest#cancel! when not cancelled sets the cancel_reason
     Failure/Error:
       expect {
         request.cancel!('testing')
       }.to change {
         request.reload.cancel_reason
       }.to eql('what')

       expected result to have changed to eql "what", but is now "testing"
     # ./spec/models/bag_man_request_spec.rb:74:in `block (4 levels) in <top (required)>'
```

dazza-codes: The only advantage to this modification is first checking the initial value to make sure that it's not already the same as the final value and the use of the `change` matcher should also help to ensure that something actually changes.

malakai97:
Updated spec to use change matcher

PR #84, opened 2016-09-27
Add application semantic version with an /api-v2/version route

### LINKED ISSUE #66: Evaluate adding an app version
Might modify and use https://github.com/dpn-admin/rails-app-versioning

dazza-codes:
This provides rake tasks to manage the semantic version for the application and a new route to support returning the app-sem-ver and the API ver, e.g.

``` sh
$ curl -H 'Authorization: Token token=hathi_token' http://localhost:3000/api-v2/version
{"app_version":"2.0.0","api_version":"2"}
```

See also https://github.com/sul-dlss/app_version_tasks, esp
- https://github.com/sul-dlss/app_version_tasks#usage

This will fix #66

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,13 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
-# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
+# for example lib/tasks/capistrano.rake, and they will automatically
+# be available to Rake.
 
 require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
+
+# From https://github.com/sul-dlss/app_version_tasks
+# A default configuration works for this application.
+spec = Gem::Specification.find_by_name 'app_version_tasks'
+load "#{spec.gem_dir}/lib/tasks/app_version_tasks.rake"
+require 'app_version_tasks'
--- a/app/controllers/version_controller.rb
+++ b/app/controllers/version_controller.rb
@@ -1,0 +1,17 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
+# Returns DPN::Server::Application::VERSION
+class VersionController < ApplicationController
+  include Authenticate
+
+  def show
+    versions = {
+      app_version: DPN::Server::Application::VERSION,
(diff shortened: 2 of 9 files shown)


Pcolar:
This is a great feature for checking version compatibility. We will need a policy on who should bump.
Good examples in the repo:
version:bump:major 
                   - bump and commit the major version (x in x.0.0)
                   - resets the minor and patch versions to zero
                   - does not push the local workspace to 'origin'
version:bump:minor - bump and commit the minor version (y in x.y.0)
                   - resets the patch version to zero
                   - does not push the local workspace to 'origin'
version:bump:patch - bump and commit the patch version (z in x.y.z)
                   - does not push the local workspace to 'origin'
version:current    - the current version
version:release    - tag and push the current version
                   - pushes to 'origin/{current_branch}'

Rakefile
```diff
@@ -1,6 +1,17 @@
...
+# for example lib/tasks/capistrano.rake, and they will automatically
+# be available to Rake.
 
 require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
+
+spec = Gem::Specification.find_by_name 'app_version_tasks'
+load "#{spec.gem_dir}/lib/tasks/app_version_tasks.rake"
+require 'app_version_tasks'
```
malakai97, line 13: Commented code here

app/controllers/version_controller.rb
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
```
malakai97, line 1: So my institution requires me to insert these headers on the files I write, but I didn't write this file.  So you're free to not include it.

malakai97 (approved):
See comments.

app/controllers/version_controller.rb
dazza-codes: At SUL-DLSS, the LICENSE at the root of the project is assumed to cover all original code in the project.  AFAIK, there are no additional requirements for headers in each file (at Stanford).

dazza-codes:
Before a merge, I'd like to do a squash on the commits; just leaving them separate for review purposes.

Rakefile
dazza-codes: Removed commented code.

dazza-codes:
Assuming a squash on this will be OK, so proceeding to do that before the end of day here in PDT.

Issue #86, opened 2016-09-27
Migration sets message_digests node_id to allow null


malakai97:
It's unclear when this was introduced, but running the migrations on a fresh database will illustrate the problem in db/schema.rb

Solutions:
1. Discover the migration that changed, and change it again.
2. Create a new migration that sets a default value to the field, disables null values, then removes the default value.  

I favor the latter.

dazza-codes:
Noticed in https://github.com/dpn-admin/dpn-server/pull/77#pullrequestreview-1623712

dazza-codes:
Checking some details related to https://github.com/dpn-admin/dpn-server/commit/cbd6e77693cdd3547722e68fbbd96297c977168e, which adds a foreign key constraint.  This might generate a different `db/schema.rb` for sqlite3 vs. MySQL.

dazza-codes:
MySQL creates a very different `db/schema.rb` file that contains the foreign key constraints, which seem to be missing from the sqlite3 version, e.g.

``` ruby
  create_table "message_digests", force: :cascade do |t|
    t.integer  "bag_id",        limit: 4,     null: false
    t.integer  "fixity_alg_id", limit: 4,     null: false
    t.text     "value",         limit: 65535, null: false
    t.integer  "node_id",       limit: 4
    t.datetime "created_at",                  null: false
  end

  add_index "message_digests", ["bag_id", "fixity_alg_id"], name: "index_message_digests_on_bag_id_and_fixity_alg_id", unique: true, using: :btree
  add_index "message_digests", ["fixity_alg_id"], name: "fk_rails_592f5598b9", using: :btree
  add_index "message_digests", ["node_id"], name: "message_digests_nodes_id_fk", using: :btree
```

Even so, the `node_id` is not set `null: false`.  This might be explained by a failure to run the migration on some existing data that does not relate to a node?  (I don't know.)

PR #88, opened 2016-09-27
Add migration to ensure MessageDigest.node_id is not null

dazza-codes:
Fix https://github.com/dpn-admin/dpn-server/issues/86


malakai97:
This will coerce all records, even those that have values.  Perhaps:
`MessageDigest.where(node_id: nil).all.each`

PR #90, opened 2016-09-29
Specs to clarify coupling between ReplicationTransfer and BagManRequest

### LINKED ISSUE #83: Explicit #cancel! specs for ReplicationTransfer, BagManRequest
Extracted from #77, see
- https://github.com/dpn-admin/dpn-server/pull/77#discussion_r80577080
- https://github.com/dpn-admin/dpn-server/pull/77#discussion_r80580539
- https://github.com/dpn-admin/dpn-server/pull/77#discussion_r80581526
- https://github.com/dpn-admin/dpn-server/pull/77#discussion_r80584879

Some of the specs could be masking the coupling between the models.  For example, the `spec/models/bag_man_request_spec.rb` has some mocks that replace the original calls, like

``` ruby
expect(replication).to receive(:cancel!)
```

This effectively makes `replication.cancel!` a non-op.  If it were changed so that it calls the original method, we might see rspec failures related to the number of times the `cancel!` methods are called in the models. So, if it were:

``` ruby
expect(replication).to receive(:cancel!).and_call_original
```

then the `Replication#cancel!` method would try to call (back to) the `BagManRequest#cancel!` method.  That should amount to 2 calls, one in the spec and another loop-back call.  The 2nd loop-back call should not execute any db changes, so long as `BagManRequest#cancelled` returns `true`.

Similar masking is present in the `replication_transfer_spec`, which could be made explicit with something like:

``` ruby
expect(replication.bag_man_request).to receive(:cancel!).and_call_original
expect(replication).to receive(:cancel!).twice.and_call_original
replication.cancel!('other', 'detail')
```

dazza-codes:
Fix #83

### DIFF
--- a/spec/models/bag_man_request_spec.rb
+++ b/spec/models/bag_man_request_spec.rb
@@ -61,40 +61,49 @@ describe BagManRequest, type: :model do
     context "when not cancelled" do
       let(:from_node_namespace) { "fake_namespace" }
       let(:replication) { Fabricate(:replication_transfer, from_node: Fabricate(:node, namespace: from_node_namespace)) }
-      let(:request) {
-        request = Fabricate(:bag_man_request)
-        allow(request).to receive(:replication_transfer).and_return(replication)
-        request
-      }
+      let(:request) { Fabricate(:bag_man_request) }
+      let(:reason) { 'testing' }
+
       before(:each) do
+        allow(request).to receive(:replication_transfer).and_return(replication)
+        allow(replication).to receive(:bag_man_request).and_return(request)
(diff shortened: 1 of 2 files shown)


Pcolar:
I'm glad to see the clarification in the code. We'll need to document thoroughly, as the sequence is a bit convoluted.  Looks OK, but I should not be the one to approve the code changes.

smutniak:
I don't notice anything wrong or missing with replication_transfer_spec.rb but am still digging through some code to make sure I understand bag_man_request properly to comment.

smutniak:
never mind

PR #91, opened 2016-09-30
Gemfile - move app_version_tasks to dev/test groups

dazza-codes:
This gem is not required in production.  During deployment to SDR dpn-dev, the git gem failed to install.  That prompted reconsideration of whether this is required in production and it should not be necessary.  The files it creates and manages are required in the app, but the gem itself is not required.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,8 +18,6 @@ end
 
 gem 'rails', '~> 4.2'
 
-gem 'app_version_tasks' # application semantic version
-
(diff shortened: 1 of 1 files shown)


smutniak:
OK.

dazza-codes:
It was closed without merging it, by mistake, so doing the merge now.

smutniak:
Yes.  Sorry about that.

PR #92, opened 2016-09-30
Rakefile - suppress failure to load app_version_tasks in production

dazza-codes:
This will fix a bug in `rake` for the production environment when it cannot load the app_version_tasks gem.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -6,8 +6,11 @@ require File.expand_path('../config/application', __FILE__)
 
 Rails.application.load_tasks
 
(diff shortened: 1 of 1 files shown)


dazza-codes:
This fixes a production deployment bug, i.e.

```
rake aborted!
Gem::LoadError: Could not find 'app_version_tasks' (>= 0) among 102 total gem(s)
/opt/app/dpn/dpn-server/releases/20160930220339/Rakefile:11:in `<top (required)>'
```

PR #105, opened 2016-10-10
Fix/temporal scopes

### LINKED ISSUE #94: Fixity/Ingest/Digest created_after and created_before are reversed
Pulled from `app/models/fixity_check.rb`

``` ruby
  scope :created_after, ->(time) { where("created_at < ?", time) unless time.blank? }
  scope :created_before, ->(time) { where("created_at > ?", time) unless time.blank? }
```

We can see `:created_after` will return items before `time` and `:created_before` will do the opposite.

We can compare to `app/models/bag.rb` as a sanity check as well (altered to match the ordering above)

``` ruby
  scope :updated_after, ->(time) { where("updated_at > ?", time) unless time.blank? }
  scope :updated_before, ->(time) { where("updated_at < ?", time) unless time.blank? }
```

malakai97:
Resolves #94

### DIFF
--- a/spec/fabricators/fixity_check_fabricator.rb
+++ b/spec/fabricators/fixity_check_fabricator.rb
@@ -9,6 +9,8 @@ Fabricator(:fixity_check) do
   bag
   node
   success true
-  fixity_at 3.seconds.ago
   created_at 1.second.ago
+
+  # fixity_at must be at the same time or before created_at
+  fixity_at {|attrs| (attrs[:created_at] || 1.second.ago) - 3.seconds}
 end
--- a/app/models/fixity_check.rb
+++ b/app/models/fixity_check.rb
@@ -33,8 +33,8 @@ class FixityCheck < ActiveRecord::Base
   validate :fixity_at_less_than_or_equal_to_created_at
 
   ### Scopes
-  scope :created_after, ->(time) { where("created_at < ?", time) unless time.blank? }
-  scope :created_before, ->(time) { where("created_at > ?", time) unless time.blank? }
+  scope :created_after, ->(time) { where("created_at > ?", time) unless time.blank? }
+  scope :created_before, ->(time) { where("created_at < ?", time) unless time.blank? }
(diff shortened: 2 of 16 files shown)


spec/spec_helper.rb
```diff
@@ -56,6 +56,8 @@
     mocks.verify_partial_doubles = true
   end
 
+  config.alias_it_should_behave_like_to :it_has_behavior, 'has behavior:'
+
```
dazza-codes: The conventional invocation for shared examples is simply `it_behaves_like`.  Why do we need something unconventional?

spec/fabricators/fixity_check_fabricator.rb
```diff
@@ -9,6 +9,6 @@
   bag
   node
   success true
-  fixity_at 3.seconds.ago
   created_at 1.second.ago
+  fixity_at {|attrs| (attrs[:created_at] || 1.second.ago) - 3.seconds}
 end
```
dazza-codes, line 16: As a novice reading this piece of code, it's not intuitive; please add commentary to explain the purpose of the temporal fields and the intended relationship between their values.  Add references to model validation, as necessary.

dazza-codes (changes requested):
Generally OK, although prefer to see conventional invocation for shared examples.  Also, not entirely clear why this change adds new scopes on models that did not have them before - are these scopes actually used anywhere in controllers or views?

spec/spec_helper.rb
malakai97: done

spec/fabricators/fixity_check_fabricator.rb
malakai97: addressed w/ comment

malakai97:
> ...not entirely clear why this change adds new scopes on models that did not have them before...

Added scopes to models based on the specification.  Did not want to touch controllers in this PR to complete the enabling of the spec-defined behavior.

dazza-codes (approved):
Looks good to go with these changes.

PR #106, opened 2016-10-10
Fix message digests migrations

### LINKED ISSUE #103: NodeNotNullInMessageDigest migration fails in mysql
See https://github.com/dpn-admin/dpn-server/pull/99

dazza-codes:
This set of migrations deployed OK to SDR dpn-dev.

Fix #103

### DIFF
--- a/db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
+++ b/db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
@@ -1,0 +1,14 @@
+class DisableFkMessageDigestsNodes < ActiveRecord::Migration
+
+  def up
+    remove_foreign_key :message_digests, :nodes
+  end
+
+  def down
+    add_foreign_key :message_digests, :nodes,
+      name: 'message_digests_nodes_id_fk',
+      column: 'node_id',
+      on_delete: :restrict,
+      on_update: :cascade
+  end
+end
--- a/db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
+++ b/db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
@@ -1,0 +1,14 @@
+class EnableFkMessageDigestsNodes < ActiveRecord::Migration
+
+  def up
+    add_foreign_key :message_digests, :nodes,
+      name: 'message_digests_nodes_id_fk',
+      column: 'node_id',
+      on_delete: :restrict,
+      on_update: :cascade
+  end
+
+  def down
+    remove_foreign_key :message_digests, :nodes
+  end
+end
--- a/db/migrate/20160927230135_node_not_null_in_message_digest.rb
+++ b/db/migrate/20160927230135_node_not_null_in_message_digest.rb
@@ -1,16 +1,8 @@
 class NodeNotNullInMessageDigest < ActiveRecord::Migration
-
   def change
-    remove_foreign_key :message_digests, column: :node_id
-    add_foreign_key :message_digests, :nodes,
-      column: :node_id,
-      on_delete: :restrict,
-      on_update: :cascade
-
     MessageDigest.where(node_id: nil).all.each do |md|
       md.update!(node_id: md.bag.admin_node_id)
     end
-    change_column :message_digests, :node_id, :integer, null: false
+    change_column_null :message_digests, :node_id, false
   end
-
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160927230135) do
+ActiveRecord::Schema.define(version: 20160927230136) do
 
   create_table "bag_man_requests", force: :cascade do |t|
     t.string   "source_location",                         null: false


dazza-codes:
After these migrations, the table creation info is:

``` sql
CREATE TABLE `message_digests` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `bag_id` int(11) NOT NULL,
  `fixity_alg_id` int(11) NOT NULL,
  `value` text COLLATE utf8_unicode_ci NOT NULL,
  `node_id` int(11) NOT NULL,
  `created_at` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_message_digests_on_bag_id_and_fixity_alg_id` (`bag_id`,`fixity_alg_id`),
  KEY `fk_rails_592f5598b9` (`fixity_alg_id`) USING BTREE,
  KEY `message_digests_nodes_id_fk` (`node_id`),
  CONSTRAINT `message_digests_nodes_id_fk` FOREIGN KEY (`node_id`) REFERENCES `nodes` (`id`) ON UPDATE CASCADE,
  CONSTRAINT `fk_rails_592f5598b9` FOREIGN KEY (`fixity_alg_id`) REFERENCES `fixity_algs` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `fk_rails_dc8eb81259` FOREIGN KEY (`bag_id`) REFERENCES `bags` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB AUTO_INCREMENT=6 DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
```diff
@@ -0,0 +1,21 @@
+class DisableFkMessageDigestsNodes < ActiveRecord::Migration
+
+  class Bag < ActiveRecord::Base
```
malakai97: These are not needed.  The migration command operates directly on the database adapter.

db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
```diff
@@ -0,0 +1,21 @@
...
+    belongs_to :bag
+  end
+
+  def up
+    remove_foreign_key :message_digests, :nodes
+  end
+
+  def down
+    add_foreign_key :message_digests, :nodes,
+      name: 'message_digests_nodes_id_fk',
```
malakai97, line 9: Should not use a specific name here, as we never set a specific name in a previous migration.

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
```diff
@@ -0,0 +1,21 @@
+class EnableFkMessageDigestsNodes < ActiveRecord::Migration
+
+  class Bag < ActiveRecord::Base
```
malakai97: Not needed, as before.

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  class MessageDigest < ActiveRecord::Base
+    belongs_to :bag
+  end
+
+  def up
+    add_foreign_key :message_digests, :nodes,
+      name: 'message_digests_nodes_id_fk',
```
malakai97, line 5: Using a specific name here doesn't break anything, but we may need to always refer to this FK by this name in the future instead of by its definition.

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
```diff
@@ -0,0 +1,21 @@
+class EnableFkMessageDigestsNodes < ActiveRecord::Migration
+
+  class Bag < ActiveRecord::Base
+  end
+
+  class MessageDigest < ActiveRecord::Base
+    belongs_to :bag
+  end
+
+  def up
```
malakai97, line 3: Given a starting state of a mysql database that has successfully run the migrations currently in develop, this migration produces an incorrect state (duplicate foreign keys):

```
  add_foreign_key "message_digests", "nodes", name: "message_digests_nodes_id_fk", on_update: :cascade
  add_foreign_key "message_digests", "nodes", on_update: :cascade
```

I imagine removing the specific name for this FK will then lead to an error about duplicate constraint names.

malakai97 (changes requested):
This PR has three migrations that are in essence trying to effect one change, so unless there's a reason they can't be put into one migration they should be.

I'd tried to say this before but I wasn't clear at all, but perhaps modifying (once again) the original offender is the best route, see https://github.com/dpn-admin/dpn-server/blob/fix/103/db/migrate/20160927230135_node_not_null_in_message_digest.rb (note this is the result of modifications to this branch that occurred after this PR was merged)

db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
dazza-codes: The migration has failed before because of conflicts in names, so this was added to fix that and it will be more robust and more meaningful with the name.

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
dazza-codes: A previous migration removes the foreign key.  The FK is the reason setting the field NOT NULL is not working, so it has to be removed and then added back.  Using a name for it makes this manipulation more reliable.

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
dazza-codes: That's the idea.  Having a name gives a concrete handle on the thing.

dazza-codes:
We have a difference of opinion about the migration being split into separate migrations or one migration.  I've found that separate migrations helps to isolate points of failure and makes for cleaner up/down code.

dazza-codes:
Removed model stubs.

db/migrate/20160927230134_disable_fk_message_digests_nodes.rb
dazza-codes: Done (for all of them).

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
dazza-codes: Done.

dazza-codes:
All of these separate migrations can rollback, i.e.

```
$ RAILS_ENV=production bundle exec rake db:migrate:status
  # ... snipped ...
   up     20160811162500  Change member uuid to member
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
   up     20160927230134  Disable fk message digests nodes
   up     20160927230135  Node not null in message digest
   up     20160927230136  Enable fk message digests nodes

$ RAILS_ENV=production bundle exec rake db:rollback STEP=1
== 20160927230136 EnableFkMessageDigestsNodes: reverting ======================
-- remove_foreign_key(:message_digests, :nodes)
   -> 0.1098s
== 20160927230136 EnableFkMessageDigestsNodes: reverted (0.1099s) =============

$ RAILS_ENV=production bundle exec rake db:rollback STEP=1
== 20160927230135 NodeNotNullInMessageDigest: reverting =======================
-- change_column_null(:message_digests, :node_id, true)
   -> 0.0933s
== 20160927230135 NodeNotNullInMessageDigest: reverted (0.1428s) ==============

$ RAILS_ENV=production bundle exec rake db:rollback STEP=1
== 20160927230134 DisableFkMessageDigestsNodes: reverting =====================
-- add_foreign_key(:message_digests, :nodes, {:name=>"message_digests_nodes_id_fk", :column=>"node_id", :on_delete=>:restrict, :on_update=>:cascade})
   -> 0.1198s
== 20160927230134 DisableFkMessageDigestsNodes: reverted (0.1201s) ============

$ RAILS_ENV=production bundle exec rake db:migrate:status
  # ... snipped ...
   up     20160811162500  Change member uuid to member
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
  down    20160927230134  Disable fk message digests nodes
  down    20160927230135  Node not null in message digest
  down    20160927230136  Enable fk message digests nodes

```

dazza-codes:
Then they can roll-forward again, i.e.

```
$ RAILS_ENV=production bundle exec rake db:migrate
== 20160927230134 DisableFkMessageDigestsNodes: migrating =====================
-- remove_foreign_key(:message_digests, :nodes)
   -> 0.0771s
== 20160927230134 DisableFkMessageDigestsNodes: migrated (0.0774s) ============

== 20160927230135 NodeNotNullInMessageDigest: migrating =======================
-- change_column_null(:message_digests, :node_id, false)
   -> 0.0806s
== 20160927230135 NodeNotNullInMessageDigest: migrated (0.1038s) ==============

== 20160927230136 EnableFkMessageDigestsNodes: migrating ======================
-- add_foreign_key(:message_digests, :nodes, {:name=>"message_digests_nodes_id_fk", :column=>"node_id", :on_delete=>:restrict, :on_update=>:cascade})
   -> 0.1225s
== 20160927230136 EnableFkMessageDigestsNodes: migrated (0.1227s) =============
```

db/migrate/20160927230136_enable_fk_message_digests_nodes.rb
malakai97: Right, migration *134 removes the fk, but the later migration *135 is already part of develop, so anyone that has pulled develop down will not run *134 by default, which puts them in the erroneous state.

dazza-codes: I think I'm the only guinea pig trying to use migrations on an existing db on a DPN-dev node.  So we might be OK on this one.

dazza-codes: For SDR dpn-dev, I think it was the previous *135 that failed to run.

dazza-codes: Let's assume that *135 already ran, then it should be the case that *134 would run (to remove any foreign key, named or not), then *135 would not run (but the effect of having a NOT NULL node_id would be in tact), and then *136 would add the named foreign key.  Is there some reason *134 would not run, even if *135 has already run?  I think the migration just iterates over them, assuming they are idempotent up/down/change ops (or should be).

dazza-codes:
Testing whether an existing `develop` migration db could fail if it then used the migrations in this PR.  First, used `develop-mysql` branch to run all the existing `develop` migrations in a clean db.  Then rebased that branch on this one and checked the migration status, i.e.

```
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
  down    20160927230134  Disable fk message digests nodes
   up     20160927230135  Node not null in message digest
  down    20160927230136  Enable fk message digests nodes
```

Then run the migrations and checked the status again:

```
 rake db:migrate
== 20160927230134 DisableFkMessageDigestsNodes: migrating =====================
-- remove_foreign_key(:message_digests, :nodes)
   -> 0.0116s
== 20160927230134 DisableFkMessageDigestsNodes: migrated (0.0117s) ============

== 20160927230136 EnableFkMessageDigestsNodes: migrating ======================
-- add_foreign_key(:message_digests, :nodes, {:name=>"message_digests_nodes_id_fk", :column=>"node_id", :on_delete=>:restrict, :on_update=>:cascade})
   -> 0.0451s
== 20160927230136 EnableFkMessageDigestsNodes: migrated (0.0455s) =============

$ rake db:migrate:status
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
   up     20160927230134  Disable fk message digests nodes
   up     20160927230135  Node not null in message digest
   up     20160927230136  Enable fk message digests nodes
```

So, everything works out OK, it seems.  The resulting `db/schema.rb` looks OK, i.e.

``` ruby
  create_table "message_digests", force: :cascade do |t|
    t.integer  "bag_id",        limit: 4,     null: false
    t.integer  "fixity_alg_id", limit: 4,     null: false
    t.text     "value",         limit: 65535, null: false
    t.integer  "node_id",       limit: 4,     null: false
    t.datetime "created_at",                  null: false
  end

  add_index "message_digests", ["bag_id", "fixity_alg_id"], name: "index_message_digests_on_bag_id_and_fixity_alg_id", unique: true, using: :btree
  add_index "message_digests", ["fixity_alg_id"], name: "fk_rails_592f5598b9", using: :btree
  add_index "message_digests", ["node_id"], name: "message_digests_nodes_id_fk", using: :btree

  add_foreign_key "message_digests", "bags", name: "message_digests_bags_id_fk", on_update: :cascade, on_delete: :cascade
  add_foreign_key "message_digests", "fixity_algs", on_update: :cascade, on_delete: :cascade
  add_foreign_key "message_digests", "nodes", name: "message_digests_nodes_id_fk", on_update: :cascade
```

malakai97:
1. > The migration has failed before because of conflicts in names, so this was added to fix that and it will be more robust and more meaningful with the name.
   
   Giving a name in general appears to address the issue you're having with a version of mysql that isn't supported by the mysql2 gem.  At this time, that may be a good enough reason to keep it.  That said, I expect incompatibilities like this will continue to arise in the future, and recommend using a database version that is officially supported by ActiveRecord and Rails.
   
   Giving a name in this _specific_ case is masking that the *136 migration is erroneously creating a duplicate key on databases that have already applied the `develop` migrations through *135.  It is turning a loud failure into a silent one.  Whether the `name` stays or goes, let's eliminate that failure.  Adding a remove_foreign_key, with the default name, in the `up` of the *136 migration should do that.
2. As written, rolling back the migrations puts the database in an invalid state that it may never have been in.  Specifically: the node_id field will allow null values despite some production databases never allowing so.   
   
   Really the point of this migration is to converge some schemas that have diverged accidentally--it doesn't change anything on schemas currently in a good state.  So the `down` operation for this set of migrations should probably just be a no-op.

dazza-codes:
1-a.  I've requested a db update to a later version of MySQL on the SDR dpn-dev system.  In any case, the migrations in this PR work on the older MySQL and more recent versions.

1-b.  WRT, "Giving a name in this specific case is masking that the *136 migration is erroneously creating a duplicate key on databases that have already applied the develop migrations through *135. It is turning a loud failure into a silent one. Whether the name stays or goes, let's eliminate that failure."  --  I'm a bit lost, because I've already tested this scenario (noted above) and I did not find that it fails or results in an invalid state.  I do not see any failures or _duplicate_ index/FK on message_digests for these migrations.  Can you please demonstrate and note the concrete operations used to replicate the invalid state (duplicate keys)?

1-c. WRT, "Adding a remove_foreign_key, with the default name, in the up of the *136 migration should do that."  This is already done in the *134 migration, so adding this to *136 results in a failure with an error about the FK missing.

2-a.  Given the migrations in this PR, it can rollback, whereas the previous version could not.  If the node_id allows a NULL value during the migrations, it will be temporary and never result in any actual NULL values in the table (the migrations ensure the value is obtained from the bag relation anyway).

2-b. The assumption that nodes have already migrated needs to be verified and slack commentary yesterday suggests that nothing has happened yet; apart from the failure on the SDR dpn-dev node, resulting in this PR.  (In general, in principle, I agree that an existing migration that is applied already needs to be supplemented by extra migrations; so I guess that's the motivation for the objection?  But, the migrations have not happened anywhere.  Also, even if they have, as in SDR dpn-dev, this PR does a decent job of preserving the intent of the previous migration; it simply separates out the components of the existing migration, to isolate the FK management, and slightly modifies the syntax of the existing migration to use `change_column_null`.  Other than that and the use of an FK name, it preserves the migration intent and it worked on the tests noted above and on the SDR dpn-dev node.)

In another branch, I will try to use the branch in this PR to run some tests without using an FK name.  However, this is difficult to test because I don't have the same MySQL version as we have on SDR dpn-dev.  I've found some migration failures on the SDR dpn-dev node because somehow it does not remove or move some kind of auto-generated index/FK and the way to work around that is to use an explicit index/FK name.

dazza-codes:
Testing migrations without an FK name on the SDR mysqldump is failing, e.g.

```
$ rake db:migrate:status
  # snipped
   up     20160921185130  Add cancel reason detail to restore transfer
   up     20160927230134  Disable fk message digests nodes
  down    20160927230135  Node not null in message digest
  down    20160927230136  Enable fk message digests nodes

$ rake db:migrate
== 20160927230135 NodeNotNullInMessageDigest: migrating =======================
-- change_column_null(:message_digests, :node_id, false)
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Column 'node_id' cannot be NOT NULL: needed in a foreign key constraint 'dpn_development/fk_rails_d4e6ffe21c' SET NULL: ALTER TABLE `message_digests` CHANGE `node_id` `node_id` int(11) NOT NULL
```

Even after a rollback, it still fails, i.e.

```
$ rake db:rollback STEP=1
== 20160927230134 DisableFkMessageDigestsNodes: reverting =====================
-- add_foreign_key(:message_digests, :nodes, {:column=>"node_id", :on_delete=>:restrict, :on_update=>:cascade})
   -> 0.0420s
== 20160927230134 DisableFkMessageDigestsNodes: reverted (0.0421s) ============

 rake db:migrate
== 20160927230134 DisableFkMessageDigestsNodes: migrating =====================
-- remove_foreign_key(:message_digests, :nodes)
   -> 0.0073s
== 20160927230134 DisableFkMessageDigestsNodes: migrated (0.0075s) ============

== 20160927230135 NodeNotNullInMessageDigest: migrating =======================
-- change_column_null(:message_digests, :node_id, false)
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

Mysql2::Error: Column 'node_id' cannot be NOT NULL: needed in a foreign key constraint 'dpn_development/fk_rails_d4e6ffe21c' SET NULL: ALTER TABLE `message_digests` CHANGE `node_id` `node_id` int(11) NOT NULL
```

This is why using an FK name is required and working in the migrations of this PR.  Note that the FK name was introduced already in an earlier migration, see https://github.com/dpn-admin/dpn-server/blob/develop/db/migrate/20160429211033_add_nodes_to_message_digests.rb#L23

So, the conclusion of this test is that using an FK name is required.

Pcolar:
WRT migrations, we will need to demonstrate the migrations on a clone of the production systems rolling forward. That can be addressed by issues in the Acceptance Tests milestone. 

1b. - I agree that we should demonstrate this failure scenario. "It is turning a loud failure into a silent one. Whether the name stays or goes, let's eliminate that failure."

2b. - Can we make the migration step conditional? "The assumption that nodes have already migrated needs to be verified"

dazza-codes:
Looking at some more details, without using the SDR dpn-dev mysqldump.  First, a checkout of `develop-mysql` and running `rake db:migrate VERSION=20160921185130` results in a migration status:

```
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
  down    20160927230135  Node not null in message digest
```

This simulates the scenario where the last migration was never run (which is most likely true for all the production nodes and all the dev nodes, except for SDR dpn-dev).  The code in this PR modifies the last migration by adding one before it, changing it, and adding one after it.

At this point, the table schema looks like so:

``` sql
CREATE TABLE `message_digests` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `bag_id` int(11) NOT NULL,
  `fixity_alg_id` int(11) NOT NULL,
  `value` text COLLATE utf8_unicode_ci NOT NULL,
  `node_id` int(11) DEFAULT NULL,
  `created_at` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_message_digests_on_bag_id_and_fixity_alg_id` (`bag_id`,`fixity_alg_id`),
  KEY `fk_rails_592f5598b9` (`fixity_alg_id`),
  KEY `message_digests_nodes_id_fk` (`node_id`),
  CONSTRAINT `fk_rails_592f5598b9` FOREIGN KEY (`fixity_alg_id`) REFERENCES `fixity_algs` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `message_digests_bags_id_fk` FOREIGN KEY (`bag_id`) REFERENCES `bags` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `message_digests_nodes_id_fk` FOREIGN KEY (`node_id`) REFERENCES `nodes` (`id`) ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

Note how the FK have names.  So, assuming that the last migration was never run, the migrations in this PR can use an FK name to manage the removal and adding of the FK.

On the other hand, if we assume the last migration was run, we run into some problems because the name for the FK is wiped out, i.e.

``` sql
CREATE TABLE `message_digests` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `bag_id` int(11) NOT NULL,
  `fixity_alg_id` int(11) NOT NULL,
  `value` text COLLATE utf8_unicode_ci NOT NULL,
  `node_id` int(11) NOT NULL,
  `created_at` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_message_digests_on_bag_id_and_fixity_alg_id` (`bag_id`,`fixity_alg_id`),
  KEY `fk_rails_592f5598b9` (`fixity_alg_id`),
  KEY `fk_rails_9ec29585a4` (`node_id`),
  CONSTRAINT `fk_rails_592f5598b9` FOREIGN KEY (`fixity_alg_id`) REFERENCES `fixity_algs` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `fk_rails_9ec29585a4` FOREIGN KEY (`node_id`) REFERENCES `nodes` (`id`) ON UPDATE CASCADE,
  CONSTRAINT `message_digests_bags_id_fk` FOREIGN KEY (`bag_id`) REFERENCES `bags` (`id`) ON DELETE CASCADE ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

In this scenario, we are also stuck with it because the last migration on the `develop` branch cannot be rolled-back, i.e.

```
 rake db:rollback STEP=1
== 20160927230135 NodeNotNullInMessageDigest: reverting =======================
rake aborted!
StandardError: An error has occurred, all later migrations canceled:

ActiveRecord::IrreversibleMigration
```

In this scenario (which is not likely a reality on most DPN dev nodes?), when we move to the code in this PR (`git checkout fix-message-digests-migration-mysql`), we have a migration status like so (assuming the last migration on `develop` succeeded [although it did not on SDR dpn-dev]):

```
   up     20160921184937  Add cancel reason detail to replication transfer
   up     20160921185130  Add cancel reason detail to restore transfer
  down    20160927230134  Disable fk message digests nodes
   up     20160927230135  Node not null in message digest
  down    20160927230136  Enable fk message digests nodes
```

So, what happens if we try to run the migrations in this PR now?  Well, it just works, i.e.

```
$ rake db:migrate
== 20160927230134 DisableFkMessageDigestsNodes: migrating =====================
-- remove_foreign_key(:message_digests, :nodes)
   -> 0.0065s
== 20160927230134 DisableFkMessageDigestsNodes: migrated (0.0067s) ============

== 20160927230136 EnableFkMessageDigestsNodes: migrating ======================
-- add_foreign_key(:message_digests, :nodes, {:name=>"message_digests_nodes_id_fk", :column=>"node_id", :on_delete=>:restrict, :on_update=>:cascade})
   -> 0.0428s
== 20160927230136 EnableFkMessageDigestsNodes: migrated (0.0430s) =============
```

The resulting schema gets the node_id NOT NULL and we get the FK names back, i.e.

```
CREATE TABLE `message_digests` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `bag_id` int(11) NOT NULL,
  `fixity_alg_id` int(11) NOT NULL,
  `value` text COLLATE utf8_unicode_ci NOT NULL,
  `node_id` int(11) NOT NULL,
  `created_at` datetime NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `index_message_digests_on_bag_id_and_fixity_alg_id` (`bag_id`,`fixity_alg_id`),
  KEY `fk_rails_592f5598b9` (`fixity_alg_id`),
  KEY `message_digests_nodes_id_fk` (`node_id`),
  CONSTRAINT `fk_rails_592f5598b9` FOREIGN KEY (`fixity_alg_id`) REFERENCES `fixity_algs` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `message_digests_bags_id_fk` FOREIGN KEY (`bag_id`) REFERENCES `bags` (`id`) ON DELETE CASCADE ON UPDATE CASCADE,
  CONSTRAINT `message_digests_nodes_id_fk` FOREIGN KEY (`node_id`) REFERENCES `nodes` (`id`) ON UPDATE CASCADE
) ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci;
```

There are no duplicate index/FK entries in this.  Also, this is the current state of the SDR dpn-dev node after applying the migrations in this PR, so it's known to work on SDR dpn-dev.

dazza-codes:
If I use the SDR dpn-dev mysqldump data, I can create migration scenarios where the result does contain duplicate FK entries.  However, I am not concerned about that because the mysqldump data does not represent a good test case -- it was obtained while debugging some migration failures and I don't recall what code branch was deployed on the system at the time of the dump.  In other words, I can't replicate the state of the migrations in that dump, so it's not a clean migration scenario.

Issue #112, opened 2016-10-19
Need own-node fixity records for integration testing


diamondap:
We need to add fixity records to our fixture data so that each "impersonate" node has one fixity record that it created locally. That is, aptrust should have a fixity check that was calculated by aptrust, chron should have one calculated by chron, etc. 

We need this to test our DPN Sync code, which should pull from each node the FixityCheck records created _at that node._

diamondap:
Commit https://github.com/dpn-admin/dpn-server/commit/1cfbbd469c6521c41f387d5b04057dc4d14b5eaf on branch fixity-fixtures should fix this issue.

PR #113, opened 2016-10-19
Added own-node fixity checks for integration testing

diamondap:
This fixes issue https://github.com/dpn-admin/dpn-server/issues/112.

### DIFF
--- a/test/fixtures/integration/aptrust/fixity_checks.yml
+++ b/test/fixtures/integration/aptrust/fixity_checks.yml
@@ -29,3 +29,11 @@ fixity04:
   success: false
   fixity_at: 2015-09-15T17:56:03.309Z
   created_at: 2015-09-15T17:56:03.309Z
+
+fixity05:
+  fixity_check_id: 99be34a2-ea07-4b49-b5fc-0da10255ac76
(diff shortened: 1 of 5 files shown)


dazza-codes:
@diamondap - the code changes here seem to fix #112 and they look simple enough, but I don’t fully understand the workflows that should create a fixity_check record.  Hence, I don’t know whether creating these records as a fixture could interfere with some other integration test or workflow.  That is, if another integration test that should create the fixity_check record is run, but the record already exists and it is read-only, then it will fail.  So, there are several facets to the integration tests - one is just replicating the registry data (like a backup of registry data across nodes), another is creating the records during some bag replication workflow.  Is it possible that creating this record in a fixture could interfere with a bag-replication workflow?

diamondap:
@darrenleeweber - As far as I understand, each node has to periodically calculate a fixity check on its copy of a bag, then record the result of that fixity check. So a single bag can have multiple fixity checks from the same node, using the same algorithm. I don't think the new fixtures will interfere with testing any of the ingest or replication workflows.

malakai97:
There shouldn't be any interference.

PR #115, opened 2016-10-19
Fixed a bug where adapter#map_bool would coerce nil values to false.

### LINKED ISSUE #114: Adapter#map_bool creates missing params - bug in /ingest
The `params[:ingested]` is set to `false` by default, it seems, so a generic call to `/ingest` is always going to return only `false` ingest records?  Is that a “feature”?

https://github.com/dpn-admin/dpn-server/blob/develop/app/adapters/ingest_adapter.rb#L12 seems to ensure that when `params[:ingested] == nil`, it becomes `false`.  In the general case where the raw HTTP params has no key for `ingested`, it will return `nil` for `params[:ingested]`.  Maybe `map_bool` needs to first check `params.keys.include?` before assigning a value?  See
- https://github.com/dpn-admin/dpn-server/blob/develop/app/adapters/adapter.rb#L31-L42

When there is some `ingest` fixture data loaded, where the record has `ingested: true`, e.g.

```
$ node=sdr
$ RAILS_ENV="impersonate_$node" IMPERSONATE=$node bundle exec rails c
Loading impersonate_sdr environment (Rails 4.2.7.1)
irb(main):001:0> Ingest.all
  Ingest Load (0.4ms)  SELECT `ingests`.* FROM `ingests`
=> #<ActiveRecord::Relation [#<Ingest id: 298216875, ingest_id: "51cae6e3-63d1-41a3-b4dc-7d021a918d6f", bag_id: 368674721, ingested: true, created_at: "2015-09-15 17:56:03">]>
```

The REST API call is not returning it, e.g.

```
$ node=sdr
$ RAILS_ENV="impersonate_$node" IMPERSONATE=$node bundle exec rails s
$ curl -L -H 'Authorization: Token token=sdr_token' http://localhost:3000/api-v2/ingest
{"count":0,"next":null,"previous":null,"results":[]}
```

So, this is handled in
https://github.com/dpn-admin/dpn-server/blob/develop/app/controllers/ingests_controller.rb#L15-L24

Some debugging indicates that it fails to find the record because `params[:ingested]` is set to `false`, even when that param is not given or required.

malakai97:
Resolves #114

### DIFF
--- a/app/adapters/adapter.rb
+++ b/app/adapters/adapter.rb
@@ -33,7 +33,11 @@ module Adapter
   # @param [Symbol] public_field
   def map_bool(model_field, public_field)
     map_from_public public_field do |value|
-      {model_field => to_bool(value) }
+      if value.nil?
+        {model_field => nil}
+      else
+        {model_field => to_bool(value) }
+      end
     end
 
     map_to_public model_field do |value|
@@ -192,7 +196,7 @@ module Adapter
 
   def to_bool(value)
     return true if value == true || value =~ (/^(true|t|yes|y|1)$/i)
-    return false if value == false || value.blank? || value =~ (/^(false|f|no|n|0)$/i)
+    return false if value == false || value =~ (/^(false|f|no|n|0)$/i)
     raise ArgumentError.new("invalid value for boolean: \"#{value}\"")
   end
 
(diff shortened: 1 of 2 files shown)


dazza-codes (approved):
Looks good, will ship it.

PR #116, opened 2016-10-20
Fix bug where ill-configured rsync-over-ssh connections would hang instead of fail

malakai97:
Adds
- BatchMode=yes - disables password querying
- ChallengeResponseAuthentication=no - disallows other interactive auth
- ConnectTimeout=3 - Timeout for the connection

### DIFF
--- a/app/jobs/client/repl/bag_retrieval_job.rb
+++ b/app/jobs/client/repl/bag_retrieval_job.rb
@@ -20,8 +20,18 @@ module Client
 
       protected
 
+      SSH_OPTIONS = [
+        "-o BatchMode=yes",
+        "-o ConnectTimeout=3",
+        "-o ChallengeResponseAuthentication=no",
+        "-o PasswordAuthentication=no",
+        "-o UserKnownHostsFile=/dev/null",
+        "-o StrictHostKeyChecking=no",
+        "-i #{Rails.configuration.transfer_private_key}"
+      ]
+
       def perform_rsync(source_location, dest_location)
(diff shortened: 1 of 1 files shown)


app/jobs/client/repl/bag_retrieval_job.rb
```diff
@@ -20,8 +20,18 @@ def perform(request, staging_dir)
...
+        "-o StrictHostKeyChecking=no",
+        "-i #{Rails.configuration.transfer_private_key}"
+      ]
+
       def perform_rsync(source_location, dest_location)
-        options = ["-a --partial -q -k --copy-unsafe-links -e 'ssh -o PasswordAuthentication=no -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no -i #{Rails.configuration.transfer_private_key}' "]
+        options = ["-a --partial -q -k --copy-unsafe-links -e 'ssh #{SSH_OPTIONS.join(" ")}' "]
         Rsync.run(source_location, dest_location, options) do |result|
           raise "Failed to transfer" unless result.success?
         end
```
dazza-codes, line 40: The only suggestion for this is to wrap the options in a method that can memoize the result of the string join, e.g.

``` ruby
def ssh_options
  @ssh_options ||= begin
    [
      # etc.
    ].join(" ")
  end
end
```

It may not help the situation if this class/object is re-instantiated for every single time it gets used.

dazza-codes (approved):
Looks OK, just one suggestion that is not required to ship-it.

app/jobs/client/repl/bag_retrieval_job.rb
malakai97: It doesn't help in this specific situation, but that doesn't matter.  Updated.

PR #118, opened 2016-10-24
Build multiple databases in travis ci

malakai97:
This pull request adds support for building sqlite, mysql, and postgres in travis.  

Three files were created to enable this: 
- config/database.yml.travis-mysql
- config/database.yml.travis-sqlite
- config/database.yml.travis-postgres

Additionally .travis.yml was extended to move the above files into place as well as setup the build matrix.

No changes to the ruby/rails code were involved.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,11 @@ before_script:
 
 script: bundle exec rspec --tag ~broken_in_ci
 
+script:
+  - if [[ $DB = 'sqlite' ]]; then bundle exec rspec --tag ~broken_in_sqlite; fi
+  - if [[ $DB = 'mysql' ]]; then bundle exec rspec; fi
+  - if [[ $DB = 'postgres' ]]; then bundle exec rspec; fi
+
 addons:
   code_climate:
     repo_token: ce328d44d99a09d03c1dac2198c731dc8bfd22374fefc1917ce528b25a06ec4e
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,9 @@ before_install:
 
 install:
   - bundle install --jobs=3 --retry=3
+  - sqlite3 --version
+  - mysql --version
+  - psql --version
 
 before_script:
   - bundle exec rake config
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,44 @@
 language: ruby
 rvm:
   - 2.3.1
+
 branches:
   only:
     - master
     - develop
-script: bundle exec rspec --tag ~broken_in_ci
+
+services:
+  - mysql
+  - postgresql
+
+env:
+  - DB=sqlite
+  - DB=mysql
+  - DB=postgres
+
+before_install:
+  - if [[ $DB = 'sqlite' ]]; then echo "gem 'sqlite3'" > Gemfile.local; fi
+  - if [[ $DB = 'mysql' ]]; then echo "gem 'mysql2'" > Gemfile.local; fi
+  - if [[ $DB = 'postgres' ]]; then echo "gem 'pg'" > Gemfile.local; fi
+
 install:
   - bundle install --jobs=3 --retry=3
-  - bundle exec sqlite3 --version
+
+before_script:
   - bundle exec rake config
+  - cp config/database.yml.travis-$DB config/database.yml
+  - if [[ $DB = 'sqlite' ]]; then bundle exec rake db:create; fi
+  - if [[ $DB = 'mysql' ]]; then mysql -e 'CREATE DATABASE dpn;'; fi
+  - if [[ $DB = 'postgres' ]]; then psql -c 'CREATE DATABASE dpn;' -U postgres; fi
+  - bundle exec rake db:migrate
+
+script: bundle exec rspec --tag ~broken_in_ci
+
 addons:
   code_climate:
     repo_token: ce328d44d99a09d03c1dac2198c731dc8bfd22374fefc1917ce528b25a06ec4e
+
 notifications:
   slack:
-    
     secure: >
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,20 +1,44 @@
...
+  - DB=postgres
+
+before_install:
+  - if [[ $DB = 'sqlite' ]]; then echo "gem 'sqlite3'" > Gemfile.local; fi
+  - if [[ $DB = 'mysql' ]]; then echo "gem 'mysql2'" > Gemfile.local; fi
+  - if [[ $DB = 'postgres' ]]; then echo "gem 'pg'" > Gemfile.local; fi
+
 install:
   - bundle install --jobs=3 --retry=3
-  - bundle exec sqlite3 --version
```
dazza-codes, line 11: Could we keep this and add similar db `--version` calls, using a similar conditional pattern to the `before_install` block?  We've already run into some version specific problems in migrations, so we're going to want these sooner or later so we know what travis is running (unless we get that info some other way?).

.travis.yml
```diff
@@ -1,20 +1,44 @@
...
+before_script:
   - bundle exec rake config
+  - cp config/database.yml.travis-$DB config/database.yml
+  - if [[ $DB = 'sqlite' ]]; then bundle exec rake db:create; fi
+  - if [[ $DB = 'mysql' ]]; then mysql -e 'CREATE DATABASE dpn;'; fi
+  - if [[ $DB = 'postgres' ]]; then psql -c 'CREATE DATABASE dpn;' -U postgres; fi
+  - bundle exec rake db:migrate
+
+script: bundle exec rspec --tag ~broken_in_ci
+
```
dazza-codes, line 39: The MySQL and Postgres builds will likely work without the `--tag ~broken_in_ci`.  We could add more `if $DB` clauses here to only exclude stuff for sqlite3 or we could change that tag so it's `broken_in_sqlite`?

config/database.yml.travis-postgres
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2015 The Regents of the University of Michigan.
+# All Rights Reserved.
+# Licensed according to the terms of the Revised BSD License
+# See LICENSE.md for details.
+
+test: 
+  adapter: postgresql
+  database: dpn
```
dazza-codes, line 8: Seems too easy - don't need to specify a user and passwd?

.travis.yml
```diff
@@ -1,20 +1,44 @@
...
-script: bundle exec rspec --tag ~broken_in_ci
+
+services:
+  - mysql
+  - postgresql
+
+env:
+  - DB=sqlite
+  - DB=mysql
+  - DB=postgres
```
dazza-codes, line 17: This bit is confusing.  It seems to set them all and I'm not sure how this works out, maybe the last one wins or something?

.travis.yml
```diff
@@ -1,20 +1,44 @@
...
+
 branches:
   only:
     - master
     - develop
-script: bundle exec rspec --tag ~broken_in_ci
+
+services:
+  - mysql
+  - postgresql
```
dazza-codes, line 12: Can this also use the `env` value for `$DB` to define the service required?

.travis.yml
```diff
@@ -1,20 +1,44 @@
...
 install:
   - bundle install --jobs=3 --retry=3
-  - bundle exec sqlite3 --version
+
+before_script:
   - bundle exec rake config
+  - cp config/database.yml.travis-$DB config/database.yml
+  - if [[ $DB = 'sqlite' ]]; then bundle exec rake db:create; fi
+  - if [[ $DB = 'mysql' ]]; then mysql -e 'CREATE DATABASE dpn;'; fi
+  - if [[ $DB = 'postgres' ]]; then psql -c 'CREATE DATABASE dpn;' -U postgres; fi
```
dazza-codes, line 35: Wondering if this `psql` needs to set the owner of the db?  On my laptop, I had to use something like

```
createdb -T template0 --owner=dpnAdmin dpn_development
```

Maybe that's only required when the user is not `postgres`.

dazza-codes (commented):
Looks pretty good, just a few questions on how things work and whether we can keep the db `--version` stuff.  Also, would be good to drop the `broken_in_ci` tag for MySQL and Postgres.

config/database.yml.travis-postgres
malakai97: https://docs.travis-ci.com/user/database-setup/#PostgreSQL

.travis.yml
malakai97: If there is, it's not documented.  I tried a couple of the obvious things and couldn't get it to work.

.travis.yml
malakai97: https://docs.travis-ci.com/user/customizing-the-build

.travis.yml
malakai97: /shrug I just did what the documentation said to do: https://docs.travis-ci.com/user/database-setup/#PostgreSQL

.travis.yml
malakai97: Indeed; I completely forgot that I removed this.

.travis.yml
malakai97: I want to be very careful about the complexity in the ci server leaking into the codebase.  Having the extra database.yml files was a necessary evil I couldn't manage to workaround.  So restricting `broken_in_ci` to `broken_in_sqlite` is great, but I don't want to one day also have `broken_in_mysql`, `broken_in_postgres`, `broken_in_mongo`, `broken_on_tuesdays`, `broken_in_linux`, etc.

dazza-codes: OK, let's just go with `broken` ;-)

dazza-codes:
Some postgres errors in https://travis-ci.org/dpn-admin/dpn-server/jobs/170304591 that might be ActiveRecord bugs or might require restructuring the AR method calls?

Not sure if this is actually a blocker for this PR, because the intent of this PR is actually fulfilled by running the Postgres tests.  Do we need to fix this stuff for this PR?

dazza-codes:
I'm going to make a call on this and say this PR is doing what it should - it builds against all 3 db platforms intended.

Issue #110, opened 2016-10-12
Test Database Migration and Rollback


Pcolar:
Based on the current production deployment, create a test instance that rolls a database instance forward from v1.x to v2.00. The test environment should also demonstrate rollback capability.

If rollback cannot be achieved for specific migration steps, document and test possible recovery scenarios.

dazza-codes:
I have a production MySQL dump from TDR for testing the API v2 migrations.

dazza-codes:
When the TDR prod dump is loaded and the migration status is checked, we get a clean status report that everything is migrated up to and including `20151215211317` migration, i.e.

```
   up     20151003181949  Remove restore statuses table
   up     20151019220704  Remove replication statuses table
   up     20151215211317  Reverse replication transfer bag man request association
  down    20160222203129  Refactor replication transfer
  down    20160226014529  Change bag man requests
  down    20160226193341  Add unpack location to bag man requests
  down    20160226213102  Add last error to bag man requests
  down    20160429211031  Change fixity checks to digests
  down    20160429211032  Add bags to digests
  down    20160429211033  Add nodes to message digests
  down    20160429211034  Add timestamps to message digests
  down    20160510205152  Create table fixity checks
  down    20160510205153  Add bags to fixity checks
  down    20160510205154  Add nodes to fixity checks
  down    20160616191514  Create ingests
  down    20160701185802  Create run times
  down    20160803214414  Delete table delayed jobs
  down    20160805220611  Change table restore transfers
  down    20160805221503  Disallow null on replication
  down    20160805224340  Allow null on restore transfer link
  down    20160811162500  Change member uuid to member
  down    20160921184937  Add cancel reason detail to replication transfer
  down    20160921185130  Add cancel reason detail to restore transfer
  down    20160927230134  Disable fk message digests nodes
  down    20160927230135  Node not null in message digest
  down    20160927230136  Enable fk message digests nodes
```

dazza-codes:
There is at least one error in the migrations, e.g.

```
 RAILS_ENV=production bundle exec rake db:migrate
== 20160222203129 RefactorReplicationTransfer: migrating ======================
-- add_column(:replication_transfers, :store_requested, :boolean, {:null=>false, :default=>false})
   -> 0.6996s
-- add_column(:replication_transfers, :stored, :boolean, {:null=>false, :default=>false})
   -> 0.7167s
-- add_column(:replication_transfers, :cancelled, :boolean, {:null=>false, :default=>false})
[...]

PR #119, opened 2016-10-24
Develop migration fixes

dazza-codes:
Tested against:
- TDR-prod dump and Chron-prod dump for MySQL
- APTrust-prod dump on Postgres

This is a partial fix for https://github.com/dpn-admin/dpn-server/issues/110 because it tests the forward-migration path on the prod-db dumps.  The rollback migration paths are yet to be tested - they might or might not work with these fixes.

### DIFF
--- a/db/migrate/20160222203129_refactor_replication_transfer.rb
+++ b/db/migrate/20160222203129_refactor_replication_transfer.rb
@@ -59,7 +59,7 @@ class RefactorReplicationTransfer < ActiveRecord::Migration
       add_column :replication_transfers, :status, :integer, null: false, default: 0
 
       ReplicationTransfer.where(cancelled: true)
-        .update_all(status: :cancelled)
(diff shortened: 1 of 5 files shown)


dazza-codes:
The last couple of commits to `Fix db:rollback ...` were tested using Chron-prod dump on `develop-mysql-migration-fixes`.  That branch, with that dump, can do a full `db:migrate` and it can nearly get all the way back to the dump `db:migrate:status` using a `db:rollback STEP=24` but the last migration rollback fails.

This gets closer to a complete "fix" for #110.  However, just because the `db:rollback` can work, doesn't mean that it should be used or that it ensures the app is functioning OK after each rollback.  Without corresponding code changes, the app will not function as expected.  The ability to rollback is only desirable if some migration failures could be prevented using the rollback.  During testing, if the migrations fail, often the database is left in an inconsistent state that requires restoring it from a backup or manually fixing it.  The rollback functionality is not a fail-safe.

PR #120, opened 2016-10-24
Minimum migration changes for api-v1 -> api-v2

malakai97:
### Goal

Aims to be a minimal version of #119.  
### Differences
- Fixes replication transfer rollback.
- Does NOT add a guard clause to prevent removing a foreign key that does not exist in the change_fixity_checks_to_digests migration.  Reasoning below.
- Does NOT separate the setting of the NOT NULL clause on message_digests.created_at into a separate migration.
- Adds full rollback definition for the delete_table_delayed_jobs migration.
- Adds slightly more rollback information for the change_table_restore_transfers migration.
### I like #119 more.  What do I need from this?
- Add this PR's fix to replication transfer rollback (6bce8f71cd6af72ae19ef7890a8e782e1cf144d3)
- Add this PR's support for rolling back the delayed job table's deletion (0608f1633d114b3cd84fcb650a865317aef7d17f)
- Remove #119's missing FK workaround (d5e4ec7305776e2d99becbaec068cd63314acf87)
### Problem with the workaround

Regarding 9ffde36c02377488a7ada1d2542fe0acab792edb " Fix migration: 20160429211031_change_fixity_checks_to_digests"
This fix only addresses a symptom of the underlying problem.  Evidently:

$ grep -i foreign mysqldump-dpn-prod-tdr-20161017.sql | grep -v FOREIGN_KEY_CHECKS | wc -l
0
This tells us the tdr database was not correctly deployed; likely it was based off the sqlite db/schema.rb which does not contain any foreign key commands.  The fix is to add the missing foreign keys to the deployed code.  If we make the change in the commit we end up hiding the underlying issue until the next time we edit a foreign key they don't have; plus the referential integrity will still be missing.

### DIFF
--- a/db/migrate/20160222203129_refactor_replication_transfer.rb
+++ b/db/migrate/20160222203129_refactor_replication_transfer.rb
@@ -59,7 +59,7 @@ class RefactorReplicationTransfer < ActiveRecord::Migration
       add_column :replication_transfers, :status, :integer, null: false, default: 0
 
       ReplicationTransfer.where(cancelled: true)
-        .update_all(status: :cancelled)
+        .update_all(status: ReplicationTransfer.statuses[:cancelled])
 
       ReplicationTransfer.where(cancel_reason: 'fixity_reject')
         .update_all(fixity_accept: false)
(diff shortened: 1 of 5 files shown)


dazza-codes:
Let's rebase on `develop` now it has travis enabled to run all 3 db platforms.

dazza-codes:
Merged this branch into #119 to consolidate this work and the PR review there.

PR #123, opened 2016-10-25
Fix latest only scope

malakai97:
The latest_only scope for Ingest and FixityCheck was previously taking advantage of non-SQL compliant behavior of sqlite and mysql.  This PR changes those two scopes to not rely on that behavior, and should thus work in all database backends.

Unfortunately, ActiveRecord has little to no support for subqueries, so the AR query involves a fair bit of manual SQL.

### DIFF
--- a/app/models/fixity_check.rb
+++ b/app/models/fixity_check.rb
@@ -38,7 +38,21 @@ class FixityCheck < ActiveRecord::Base
   scope :with_success, ->(success) { where(success: success) unless success.blank? }
   scope :with_bag_id, ->(id) { where(bag_id: id) unless id.blank? }
   scope :with_node_id, ->(id) { where(node_id: id) unless id.blank? }
-  scope :latest_only, ->(flag) { group(:node_id, :bag_id).having("created_at = max(created_at)") unless flag.blank? }
+  scope :latest_only, ->(flag) do
+    unless flag.blank?
+      joins("INNER JOIN (
+        SELECT node_id, bag_id, MAX(created_at) AS max_created_at
+        FROM #{FixityCheck.table_name}
+        GROUP BY node_id, bag_id
+        ) AS x
+        ON
+        #{FixityCheck.table_name}.bag_id = x.bag_id
+        AND
+        #{FixityCheck.table_name}.node_id = x.node_id
+        AND
+        #{FixityCheck.table_name}.created_at = x.max_created_at")
+    end
+  end
 
   private
   def fixity_at_less_than_or_equal_to_created_at
--- a/app/models/ingest.rb
+++ b/app/models/ingest.rb
@@ -35,8 +35,20 @@ class Ingest < ActiveRecord::Base
   scope :created_before, ->(time) { where("created_at < ?", time) unless time.blank? }
   scope :with_bag_id, ->(id) { where(bag_id: id) unless id.blank? }
   scope :with_ingested, ->(ingested) { where(ingested: ingested) if [true,false].include?(ingested) }
-  scope :latest_only, ->(flag) { group(:bag_id).having("created_at = max(created_at)") unless flag.blank? }
-  
+  scope :latest_only, ->(flag) do
+    unless flag.blank?
+      joins("INNER JOIN (
+        SELECT bag_id, MAX(created_at) AS max_created_at
+        FROM #{Ingest.table_name}
+        GROUP BY bag_id
+        ) AS x
+        ON
+        #{Ingest.table_name}.bag_id = x.bag_id
+        AND
+        #{Ingest.table_name}.created_at = x.max_created_at")
+    end
+  end
+
   private
   
   def fail_unless_new(node)
@@ -47,3 +59,5 @@ class Ingest < ActiveRecord::Base
   end
 
 end
+
+
--- a/spec/models/fixity_check_spec.rb
+++ b/spec/models/fixity_check_spec.rb
@@ -55,7 +55,7 @@ describe FixityCheck do
 
   it_behaves_like "it has temporal scopes for", :created_at
 
-  describe "scope latest_only", :broken_in_sqlite do
+  describe "scope latest_only" do
     before(:each) do
       @bag1, @bag2 = Fabricate.times(2, :bag)
       @node1, @node2 = Fabricate.times(2, :node)
--- a/spec/models/ingest_spec.rb
+++ b/spec/models/ingest_spec.rb
@@ -100,7 +100,7 @@ describe Ingest do
     end
   end
 
-  describe "scope latest_only", :broken_in_sqlite do
+  describe "scope latest_only" do
     before(:each) do
       @bag1 = Fabricate(:bag)
       @bag2 = Fabricate(:bag)


spec/models/fixity_check_spec.rb
```diff
@@ -55,7 +55,7 @@
 
   it_behaves_like "it has temporal scopes for", :created_at
 
-  describe "scope latest_only", :broken_in_sqlite do
+  describe "scope latest_only" do
```
dazza-codes, line 58: :+1:

spec/models/ingest_spec.rb
```diff
@@ -100,7 +100,7 @@
     end
   end
 
-  describe "scope latest_only", :broken_in_sqlite do
+  describe "scope latest_only" do
```
dazza-codes, line 103: :+1:

app/models/fixity_check.rb
```diff
@@ -39,6 +39,21 @@ def self.find_fields
   scope :with_bag_id, ->(id) { where(bag_id: id) unless id.blank? }
   scope :with_node_id, ->(id) { where(node_id: id) unless id.blank? }
   scope :latest_only, ->(flag) { group(:node_id, :bag_id).having("created_at = max(created_at)") unless flag.blank? }
```
dazza-codes: This old version needs to be deleted.

app/models/fixity_check.rb
```diff
@@ -39,6 +39,21 @@ def self.find_fields
...
+        GROUP BY node_id, bag_id
+        ) AS x
+        ON
+        #{FixityCheck.table_name}.bag_id = x.bag_id
+        AND
+        #{FixityCheck.table_name}.node_id = x.node_id
+        AND
+        #{FixityCheck.table_name}.created_at = x.max_created_at")
+    end
+  end
```
dazza-codes, line 55: Wonder if there is an easier way to do this?  Not required, just offering suggestion, e.g.

``` ruby
scope :latest_only, ->(flag) do
  unless flag.blank?
    max_created_at = FixityCheck.order(:created_at).last.created_at
    records = FixityCheck.where("created_at >= ?", max_created_at).map {|r| [r.node_id, r.bag_id, r.created_at]}
    # some ruby array magic for the `GROUP BY` operation?
  end
end
```

This requires that `created_at` is indexed for better performance.

When dropping into `pry` during the specs and looking at the data, there is not a lot of variation in the timestamps of the records, e.g.

```
[22] pry(FixityCheck)> FixityCheck.order(:created_at).map {|r| r.created_at }
=> [Sun, 23 Oct 2016 22:32:54 UTC +00:00,
 Sun, 23 Oct 2016 22:32:54 UTC +00:00,
 Sun, 23 Oct 2016 22:32:54 UTC +00:00,
 Tue, 25 Oct 2016 21:32:54 UTC +00:00,
 Tue, 25 Oct 2016 21:32:54 UTC +00:00,
 Tue, 25 Oct 2016 21:32:54 UTC +00:00]
[23] pry(FixityCheck)> FixityCheck.order(:created_at).map {|r| r.created_at.to_i }
=> [1477261974, 1477261974, 1477261974, 1477431174, 1477431174, 1477431174]
```

However, there does seem to be millisecond or nanosecond variation, e.g.

``` ruby
[37] pry(FixityCheck)> FixityCheck.order(:created_at).map {|r| r.created_at.to_f }
=> [1477261974.379971, 1477261974.385619, 1477261974.390767, 1477431174.37681, 1477431174.382953, 1477431174.388226]
[38] pry(FixityCheck)> FixityCheck.order(:created_at).map {|r| r.created_at.to_f }.max
=> 1477431174.388226
```

This might explain why the following result set only has one record, even though it would appear there are 3 records with an almost identical timestamp:

``` ruby
[39] pry(FixityCheck)> FixityCheck.where("created_at >= ?", max_created_at)
=> [#<FixityCheck:0x0055e323425750
  id: 5,
  fixity_check_id: "7f49ba11-71e2-4716-a1cb-da1baa803d22",
  bag_id: 1,
  node_id: 10,
  success: true,
  fixity_at: Tue, 25 Oct 2016 20:32:54 UTC +00:00,
  created_at: Tue, 25 Oct 2016 21:32:54 UTC +00:00>]
```

It might be better if the fixture data had more variation with a known number of records, with known `id`, having the `latest` timestamp.  The details of the timestamp inequality operators seem to be non-trivial.  Perhaps the explicit SQL statements resolve any ruby data munging issues, but might leave similar concerns to the RDBMS implementation details.

app/models/ingest.rb
```diff
@@ -35,8 +35,20 @@ def self.find_fields
   scope :created_before, ->(time) { where("created_at < ?", time) unless time.blank? }
   scope :with_bag_id, ->(id) { where(bag_id: id) unless id.blank? }
   scope :with_ingested, ->(ingested) { where(ingested: ingested) if [true,false].include?(ingested) }
-  scope :latest_only, ->(flag) { group(:bag_id).having("created_at = max(created_at)") unless flag.blank? }
-  
+  scope :latest_only, ->(flag) do
+    unless flag.blank?
+      joins("INNER JOIN (
+        SELECT bag_id, MAX(created_at) AS max_created_at
```
dazza-codes, line 41: Why not just select the `created_at` value from the record instead of using the generic `max_created_at` value?  If they are all equal anyway... this might avoid having to do the `INNER JOIN`?  (It's not a financial system, so I doubt we need nanosecond resolution/accuracy from the query.)

dazza-codes: Related SO question/answer:
http://stackoverflow.com/questions/9780169/active-record-model-find-last

malakai97: They're not equal.

app/models/fixity_check.rb
malakai97: Fixed.

app/models/fixity_check.rb
malakai97: I'd rather not fetch all of the records.  Your `records =` query is not going to return the latest fixity check for each bag-node pair.

The fixture data is more varied than your query illustrates.

```
[2] pry(#<RSpec::ExampleGroups::FixityCheck::ScopeLatestOnly>)> FixityCheck.order(:bag_id, :node_id, :created_at).pluck(:bag_id, :node_id, :created_at)
=> [[1, 9, Mon, 24 Oct 2016 16:55:10 UTC +00:00],
 [1, 9, Wed, 26 Oct 2016 15:55:10 UTC +00:00],
 [1, 10, Mon, 24 Oct 2016 16:55:10 UTC +00:00],
 [1, 10, Wed, 26 Oct 2016 15:55:10 UTC +00:00],
 [2, 9, Mon, 24 Oct 2016 16:55:10 UTC +00:00],
 [2, 9, Wed, 26 Oct 2016 15:55:10 UTC +00:00]]
```

dazza-codes: OK, I get it.  Good enough.

app/models/ingest.rb
dazza-codes: OK, I was missing the point of the group-by operation and the consequent selection of the latest within each group.

dazza-codes (approved):

PR #132, opened 2016-11-02
Rails view adapters gem

### LINKED ISSUE #79: Refactor app/adapters
While looking at `app/adapters/adapter.rb` and `app/adapters/abstract_adapter.rb`, it seems they could be combined into one class, unless there is a design pattern involved? Also, it's not entirely clear why the `abstract_adapter.rb` requires both class and instance methods (it seems lines 50-57 try to create instance methods that delegate to class methods for `:simple_maps` and `:to_maps`, but they seem to be instance methods already in `adapter.rb`).

`Adapter#map_to_public` is not fully covered by unit tests.

Extracted this issue from #77

malakai97:
The DSL for the adapters has been extracted into a [gem](https://rubygems.org/gems/rails_view_adapters).

This PR adds it as a dependency, removes the now-duplicated functionality, and slightly changes the invocation to use the gem.  It removes some duplicate functionality for converting time-strings to times.  It reverts the workaround for issue #124.

The adapter tests were not modified, so we should not expect functionality to change.

Resolves #79, #124


dazza-codes:
The gem extraction is an interesting move and could work out OK, in time.  At present, it should be hooked up to travis and codeclimate to evaluate the test coverage and any other issues before moving on this PR.  There are currently no badges to indicate that this gem is tested on travis or code climate, see:

https://github.com/mlibrary/rails_view_adapters

Another concern is that this dpn-server project has seen various fixes in the adapter code in the last few months, including a currently open issue/PR to address concerns about iso8601 date-time conformance (which is specific to DPN).  The adapters and related configuration files do not yet appear to be a robust code base that can be used as a reliable gem plugin, because extracting the details to a gem will make it more difficult to debug and refactor it for dpn-server.  Also, it's not clear how a gem abstraction can be molded to deal with DPN specific details.  That is, this PR it will make it more difficult to understand and maintain or bug-fix dpn-server, which equates to more cost to maintain the system.  If the adapters in the gem become more widely adopted and receive wider testing and bug-fixing and proved to be robust and reliable, then it would be good to reconsider this approach.

dazza-codes (changes requested):
I can't approve this without first resolving concerns about `map_date` functionality and related iso8601 conformance for DPN; the subject of an open issue and PR.

malakai97:
Updated to use the latest version of rails_view_adapters, which relies on a customized version of Time.zone.strptime instead of Time.zone.parse.  This function uses DateTime._strptime at the lowest level, the big difference being it accepts a format string.

The end result is that the functionality is to set invalid datetime fields in the params to nil.  This is the same functionality that dpn-server currently uses via ApplicationController#convert_time_strings.

Also added badges.

dazza-codes:
Copying some important, related slack commentary here:

DLW:  So there are a few places to do the 'exception handling': controller/adapter while parsing the request params, the model validators and lastly  the db-schema constraints for unique values, FK constraints etc.  Although a preference for one of these is fair (i.e. Model validators), if it doesn't provide useful exception information to a client, why not do it elsewhere (adapter)?  If the request param is not an iso8601 date time value, raise and say so at the time you know it to be so.

BH:  Darren you make a good point regarding the usefulness of the exception (or other error handling behavior).  I could see validating problematic fields either ahead of the adapters or by passing more explicit closures to them.  With 132 I wanted to refactor but not change the behavior.  I can see pushing up the exception being done in another change.

dazza-codes:
There are gem updates related to this PR in https://github.com/mlibrary/rails_view_adapters.  The build and code coverage badges partially address the concern about the gem code quality.  It would help to also have code-climate analysis.

dazza-codes:
Part of the concern for this PR is the abstraction of the adapters into a gem.  There are changes in that gem that are related to functionality for this project, but they are now opaque to this project and the github review process.  A reviewer has to move to another project, where there is no related PR for that work, to try to understand the changes.  This is not a feasible process because it wastes time and obscures the relationship between bugs reported on dpn-server that are fixed elsewhere.  This substantially raises the difficulty and the costs of maintaining and extending the application.

It's therefore my view that this approach is not an optimal pathway forward for this application.  However, this is my view and does not warrant closing this PR.  I recommend additional DPN developers review this approach and discuss it's merits.  Although it seems to be a level of abstraction that will result in more costly maintenance, it is a solution that could work, so it deserves more consideration and evaluation.

malakai97:
Code Climate seems to have changed their business model; it's no longer possible for me to setup a github hook to their site with a public repo within an organization.  I've ran their stuff separately and all it's doing is running rubocop and code coverage, which this project already does--outside of a hook for the former, and with coveralls for the latter.

Pcolar:
I can extend or pay for additional capabilities in CodeClimate. I'm not sure how the account is currently configured, but will take a look.

malakai97:
Code climate was only used originally because it was easy to setup and looked interesting.  Now that we've used it, all it really provides is a shinier interface over some OSS tools.

dazza-codes:
One example of the added difficulty of working with the gem is access to contributing to it.  The gem repo is not in the DPN org, so it's not sufficient to be a DPN developer in order to make changes to it; e.g.
```
$ git push --set-upstream origin rubocop-auto-correct
ERROR: Permission to mlibrary/rails_view_adapters.git denied to darrenleeweber.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

Issue #133, opened 2016-11-02
Cannot run script/setup_cluster.rb with default Gemfile


malakai97:
`bundle exec script/setup_cluster.rb` produces the error:

```
Couldn't create database for {"adapter"=>"sqlite3", "encoding"=>"utf8", "pool"=>5, "timeout"=>5000, "reconnect"=>true, "port"=>3306, "database"=>"db/impersonate_hathi.sqlite3"}
rake aborted!
Gem::LoadError: Specified 'sqlite3' for database adapter, but the gem is not loaded. Add `gem 'sqlite3'` to your Gemfile (and ensure its version is at the minimum required by ActiveRecord).
```

This is due to the file setting the environment variable to something not known to the Gemfile/bundler before loading the bundle, so only those dependencies not restricted to `:development`, `:test`, etc. will be installed and available.

Workaround for now is to add `gem 'sqlite3'` to your Gemfile.local.

-----

### To recreate this bug

On a system with no system-wide sqlite3 gem install, i.e.:
```
$ gem list

*** LOCAL GEMS ***

bigdecimal (1.2.6)
bundler (1.13.6)
coderay (1.1.1)
io-console (0.4.3)
json (1.8.1)
method_source (0.8.2)
minitest (5.4.3)
power_assert (0.2.2)
pry (0.10.4)
psych (2.0.8)
rake (10.4.2)
rdoc (4.2.0)
slop (3.6.0)
test-unit (3.0.8)

```
Run:
```
git clone git@github.com:dpn-admin/dpn-server.git
cd dpn-server/
rbenv local 2.3.1 
git checkout develop
bundle install --path .bundle
bundle exec rake config
bundle exec ./script/setup_cluster.rb  # this generates the error
```

malakai97:
Most likely applies to script/migrate_cluster.rb as well.

dazza-codes:
Can't replicate this bug on a clean clone of the repository, using the following:

``` sh
git clone git@github.com:dpn-admin/dpn-server.git
cd dpn-server/
rbenv local 2.3.1 
git checkout develop
bundle install
bundle exec rake config
bundle exec rake db:create  # might not be necessary
bundle exec rake db:migrate  # might not be necessary
bundle exec ./script/setup_cluster.rb 
bundle exec ./script/run_cluster.rb -f
```

malakai97:
Added better steps to recreate to the opening post with your snippet.  It appears the script is picking up your system sqlite3 gem.

[...]

PR #136, opened 2016-11-02
Gemset loading - use Bundle.setup instead of Bundle.require????

dazza-codes:
Connected to #133 (for reference only, not supposed to be resolution for it).


malakai97:
I moved the pertinent information from this PR and branch to the issue itself, so this is no longer needed.  The issue isnt' under active dev, so the convenience of this pr isn't useful right now.

PR #141, opened 2016-11-14
Redesign adapters to map belongs_to associations to objects, not their IDs

### LINKED ISSUE #137: replication transfer query does not use bag UUID to restrict result set
Here's an example, using the `dpn-client` gem in a ruby console for the DPN-dev network:
```
> query
=> {:page_size=>50, :bag=>"0089f171-5cf8-4c4f-bfd4-3b6544ee8ab4", :from_node=>"chron", :to_node=>"sdr"}
> replications = []
> chron_node.dpn_client.replications(query) {|response| replications << response.body }
> replications.map {|repl| repl[:bag]}.uniq
=> ["a32e18e4-8bf4-4bc0-9e98-3f51da9ea2a1",
 "6f31995a-bf8d-4093-9f8c-b16a8f00288d",
 "1f806acb-52be-40d8-a9f9-2b6891cc62cf",
 "6b0f0fbf-f1b6-412e-b2f7-db6f35ff7154",
 "44023f57-e131-437d-be94-8d756fa2b8e5",
 "0d7ddb79-8f63-4ce3-8c90-f8452c6386fe",
 "a41f2d72-185a-4b81-bb1d-b90f907ecc7a",
 "60d28964-d18f-466a-8736-f2ceab3d320a",
 "d24e9708-6bef-432d-a1f4-df0955e79c9b"]
```

Notice how the query is returning repl-transfers with a lot of various bag-uuid values that are not the same as the bag-uuid in the query.

malakai97:
This PR resolves #137 by redesigning `Adapter#map_belongs_to` to result in a hash that contains the model object that was found, not its id.  When a model object can't be found, this PR returns a new, unpersisted model object instead of returning nil.  

This change allows the controller (and thus the models) to differentiate between a (1) a query parameter that was not provided and (2) a query parameter that filters on a record that does not exist.  

Beyond the above, the following changes were necessary to effect this change:

1. The scopes were changed to accept objects, and test whether those objects were persisted.  It is notable that scopes know how to handle `nil` by default.
2. Models needed the `validates_associated <association_name>` directive in order to fail validation when the association pointed to an invalid, unpersisted record.
3. Strong parameters in the controllers needed to be modified to explicitly include the association by name, instead of by id.
4. `Bag#update_with_associations` was modified to explicitly include the associations and not the `_id` field.


malakai97:
Currently this PR conflicts with #132, but the changes are sufficiently separate that merging this on top of that is very easy to do.

dazza-codes:
So, is this pending #132?  (If so, let's add `[Pending 132]` to the title.)

malakai97:
While it would be slightly more convenient to wait, this PR is not pending any other PRs.

PR #142, opened 2016-12-03
More robust seed functionality

malakai97:
This pull request greatly improves the power and robustness of the seed functionality we have.  It takes advantage of the seed feature provided by rails.

### Previous usage

This is unchanged from before 

* `rake db:setup` still runs the seed process, and it creates the same seeds as before.
* `rake db:seed` creates the same seeds as before, but also creates replicate/restore agreements between nodes.

### What's new

The seed operation will now load any seed files in `db/seeds/RAILS_ENV-seeds.d`, in lexical order.  It is somewhat intelligent about loading dependencies before records that rely on them.  

The seed operation now provides some output, similar to what the script in the private repo supplied.  The operation is now idempotent--it will skip existing records.

This should replace the manual adding of nodes and members that was being done previously.


malakai97:
Listed as in progress because I need to update the readme to reflect this, but ready otherwise.

dazza-codes:
Codacy indicates there are some duplicate blocks in the specs that might be consolidated into a shared_examples block, see
- https://www.codacy.com/app/DPN/dpn-server/pullRequest?prid=438899
- esp. https://www.codacy.com/app/DPN/dpn-server/file/4633489514/duplication/source?fileBranchId=3960703

db/seeds.rb
```diff
@@ -3,60 +3,12 @@
...
-        from_node.replicate_to_nodes << to_node
-        from_node.restore_to_nodes << to_node
-      end
-      from_node.save!
-    end
-  end
+seeder = Db::Seeder.new(Logger.new(STDOUT))
+Db::SeedFileLoader.new(Rails.env.to_s).seed_hashes do |hash|
+  seeder.seed!(hash)
 end
```
dazza-codes, line 12: How about `Db::Seeder.new(logger)` so it uses the same logger defined above?

db/seeds/development-seeds.d/seeds.json
```diff
@@ -0,0 +1,93 @@
...
+      "fixity_algs": ["md5","sha256"],
+      "protocols": ["rsync"],
+      "api_root": "http://127.0.0.1",
+      "private_auth_token": "hathi_token",
+      "auth_credential": "hathi_token",
+      "replicate_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "replicate_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"]
+    },
```
dazza-codes, line 31: How do these `api_root` values (for all these nodes) intersect with the scripts for the local cluster?  (The scripts simply assume there is an array of nodes and assigns a localhost:port to each of them, in turn.  So the only difference between them is the port numbers.)

db/seeds/development-seeds.d/seeds.json
```diff
@@ -0,0 +1,93 @@
...
+      "storage_type": "default",
+      "fixity_algs": ["md5","sha256"],
+      "protocols": ["rsync"],
+      "api_root": "http://127.0.0.1",
+      "private_auth_token": "hathi_token",
+      "auth_credential": "hathi_token",
+      "replicate_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "replicate_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"]
```
dazza-codes, line 30: "hathi" replicates to itself?  Does that entail an infinite loop of replication?

db/seeds/development-seeds.d/seeds.json
```diff
@@ -0,0 +1,93 @@
...
+      "storage_type": "default",
+      "fixity_algs": ["md5","sha256"],
+      "protocols": ["rsync"],
+      "api_root": "http://192.168.33.11",
+      "private_auth_token": "sdr_token",
+      "auth_credential": "sdr_token",
+      "replicate_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "replicate_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"]
```
dazza-codes, line 45: Would rather see a coding solution to setting all the `replicate_to_nodes`/`replicate_from_nodes` and the same for the restore - it looks like a simpler thing to code that to enumerate in this seeds file.  Just add a comment in the seeds file about how it is handled somewhere in code?

db/seeds/development-seeds.d/seeds.json
```diff
@@ -0,0 +1,93 @@
...
+      "fixity_algs": ["md5","sha256"],
+      "protocols": ["rsync"],
+      "api_root": "http://192.168.33.14",
+      "private_auth_token": "aptrust_token",
+      "auth_credential": "aptrust_token",
+      "replicate_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "replicate_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_to_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"],
+      "restore_from_nodes": ["hathi", "sdr", "tdr", "chron", "aptrust"]
+    }
```
dazza-codes, line 91: Most of the fields in the nodes could be handled in code, as they are nearly all identical (so long as that is how they should or must be).

db/seeds/prod.seeds.json
```diff
@@ -1,15 +0,0 @@
-{
```
dazza-codes, line 1: Hopefully the seeder throws an exception for a Rails production env - should it be impossible to use seeds on a production system?  It might be safer to disable seeds on prod.  If so, do not even keep the prod directory.

db/seeds/production-seeds.d/base_seeds.json
```diff
@@ -0,0 +1,15 @@
...
+  "protocols":[
+    { "name": "rsync" }
+  ],
+  "storage_regions": [
+    { "name": "default" }
+  ],
+  "storage_types": [
+    { "name": "default" }
+  ]
+}
```
dazza-codes: Leaning toward deleting this file and anything prod-related.

lib/db/seed_file_loader.rb
```diff
@@ -0,0 +1,32 @@
...
+  class SeedFileLoader
+    BASE_DIR = File.join "db", "seeds"
+
+    attr_reader :seed_dir
+    def initialize(env_name, base_dir = BASE_DIR)
+      @seed_dir = File.join(base_dir, "#{env_name}-seeds.d")
+    end
+
+    def seed_files
+      Dir[File.join(seed_dir, "*")].sort
```
dazza-codes: Memoize this using `@seed_files ||= ...`?

lib/db/seed_file_loader.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Db
+
+  # Loads seed files and converts to a ruby hash.
+  class SeedFileLoader
+    BASE_DIR = File.join "db", "seeds"
+
+    attr_reader :seed_dir
+    def initialize(env_name, base_dir = BASE_DIR)
+      @seed_dir = File.join(base_dir, "#{env_name}-seeds.d")
```
dazza-codes, line 17: Throw an exception if the seed directory does not exist?

lib/db/seed_file_loader.rb
```diff
@@ -0,0 +1,32 @@
...
+    end
+
+    def seed_files
+      Dir[File.join(seed_dir, "*")].sort
+    end
+
+    def seed_hashes
+      seed_files.each do |f|
+        seed_hash = JSON.parse(File.read(f), symbolize_names: true)
+        yield seed_hash
```
dazza-codes, line 27: Maybe `yield JSON.load(f, symbolize_names: true)` would do it?

lib/db/seed_file_loader.rb
dazza-codes: Also, is the order of the files important?  If so, document the files expected to be present in the directory and, perhaps, validate they are present in this method?

lib/db/seeder.rb
```diff
@@ -0,0 +1,200 @@
...
+    #   @option entries [String] :storage_region StorageRegion name
+    #   @option entries [String] :storage_type StorageType name
+    #   @option entries [Array<String>] :fixity_algs FixityAlg names
+    #   @option entries [Array<String>] :protocols Protocol names
+    #   @option entries [String] :api_root
+    #   @option entries [String] :private_auth_token Unencrypted
+    #   @option entries [String] :auth_credential Unencrypted
+    def create(type, entries)
+      create_from_entries(entries, *maps[type])
+    end
```
dazza-codes, line 61: The documentation for this is suggesting that the seed files require json-schema validations [1,2].  But, that's overkill.  Given that we know everything there is to know about most of these seed data, why not move the details from the json file into code directly?  It could be just as easy to change the code as to change the json files, if they ever need to be changed.  That is, all this `maps` apparatus is looking kinda scary for handling what it really handles.

[1] http://json-schema.org/
[2] https://github.com/ruby-json-schema/json-schema

lib/db/seed_file_loader.rb
malakai97: The case where there are no seed files is not exceptional.  Currently, seed_files on a non-extant dir will return `[]`, and this code properly does nothing.

db/seeds/production-seeds.d/base_seeds.json
malakai97: Indeed, dpn-admin/private is a better place for this.

lib/db/seeder.rb
dazza-codes: In this regard, what about just getting the json files right (maybe doing json-schema validation, if necessary) and then just using http://apidock.com/rails/ActiveRecord/Serialization/from_json ?

db/seeds/prod.seeds.json
malakai97: No, it definitely should.  That is what these changes are for.

db/seeds/development-seeds.d/seeds.json
malakai97: The code is now a general solution, so we can handle this configuration in data.

db/seeds/development-seeds.d/seeds.json
malakai97: It implies a cyclical graph, but I don't see why you'd walk it forever.  Self-replication is supported in the spec following a need from Texas.

db/seeds/development-seeds.d/seeds.json
malakai97: Unknown, but identically to as before this change.

lib/db/seeder.rb
```diff
@@ -0,0 +1,200 @@
...
+        name: node_entry[:name],
+        storage_region: StorageRegion.find_by_name(node_entry[:storage_region]),
+        storage_type: StorageType.find_by_name(node_entry[:storage_type]),
+        fixity_algs: FixityAlg.where(name: node_entry[:fixity_algs]),
+        protocols: Protocol.where(name: node_entry[:protocols]),
+        api_root: node_entry[:api_root],
+        private_auth_token: node_entry[:private_auth_token],
+        auth_credential: node_entry[:auth_credential]
+      }
+    end
```
dazza-codes, line 193: Doesn't the model do all the entry -> Model-instance translations in a before save callback or something?  Why does this code here need to do all these `find_by_name` calls?  Surely the `Node` model can figure this all out?  Suppose this were simply `Node.new.from_json node_entry` - why doesn't that work?  If it doesn't work, can we make it work?

db/seeds.rb
malakai97: Originally I +1'd this, but then I realized that the SeedFileLoader does not log anything.  I suppose it could, but I don't see the utility.

lib/db/seeder.rb
malakai97: Could you elaborate?

lib/db/seeder.rb
dazza-codes: Notice how creating a `Node` instance already depends on having existing (or creating) each of the dependent model values.  This implies that the seeds json files must contain all these dependent model data and that they must be handled in a specific order.

spec/db/seed_file_loader_spec.rb
```diff
@@ -0,0 +1,76 @@
+
+# frozen_string_literal: true
+require "rails_helper"
+require "fileutils"
+require "json"
+
+TMPDIR = File.join File.dirname(__FILE__), "tmp"
```
dazza-codes, line 7: Consider https://ruby-doc.org/stdlib-2.0.0/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir

spec/db/seed_file_loader_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  let(:complex_seed) { SpecSeeds.complex }
+  let(:mixed_keys_seed) { SpecSeeds.mixed_keys }
+  let(:base_dir) { TMPDIR }
+  let(:seed_dir) { File.join base_dir, "test-seeds.d" }
+  let(:loader) { described_class.new("test", base_dir) }
+
+  describe "#seed_dir" do
+    it "constructs the correct path" do
+      expect(loader.seed_dir).to eql(seed_dir)
+    end
```
dazza-codes, line 55: Add a spec about raising an exception when the path doesn't exist?

spec/db/seed_file_loader_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        b: [2, 3, 4],
+        c: [{ ca: 1, cb: 2 }],
+        d: { da: [1, 2, 3] }
+      }
+    end
+
+    def mixed_keys
+      { "a" => 1, b: 2 }
+    end
+  end
```
dazza-codes, line 30: Some of this spec could be easier using json-schema validations.  When the json content must be just so, the json-schema option might be a better way to ensure it.

lib/db/seeder.rb
dazza-codes: An example of using json-schema can be seen in https://github.com/ncbo/ontologies_api/blob/master/test/controllers/test_projects_controller.rb

spec/db/seed_file_loader_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        File.join(seed_dir, basename)
+      end
+    end
+    it "returns the seed files in the seed_dir" do
+      expect(loader.seed_files).to match_array(seed_files)
+    end
+  end
+
+  describe "#seed_hashes" do
+    let(:symbolized_mixed_keys_seed) { { a: 1, b: 2 } }
```
dazza-codes, line 70: `let(:symbolized_mixed_keys_seed) { mixed_keys.symbolize_keys }` ?

http://apidock.com/rails/Hash/symbolize_keys

spec/db/seed_file_loader_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    it "yields each hash successively" do
+      expect {|spy| loader.seed_hashes(&spy) }
+        .to yield_successive_args(complex_seed, symbolized_mixed_keys_seed, simple_seed)
+    end
```
darrenleeweber: Nifty use of `spy` :+1:

spec/db/seeder_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+  #   in the log.
+  shared_examples "a seedable class" do |klass|
+    it "creates #{klass} records" do
+      seeder.create(type, entries)
+      stored_records_as_arrays = klass.pluck(*entries.first.keys) # We do this because pluck auto-flattens
+        .map{|val_or_vals| [val_or_vals].flatten }                # when you only pluck one field.
+      expect(stored_records_as_arrays).to include(*entries.map(&:values))
+    end
+    it "skips existing #{klass} records" do
+      Fabricate(fabricator_name, entries.first)
```
dazza-codes, line 84: The fabricators already exist, right?

spec/db/seeder_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+      let(:display_field) { :name }
+    end
+  end
+
+  describe "create members" do
+    it_behaves_like "a seedable class", Member do
+      let(:entries) do
+        2.times.map do
+          { member_id: SecureRandom.uuid, name: Faker::Company.name, email: Faker::Internet.email }
+        end
```
dazza-codes, line 132: Wondering whether a fabricator build with to_hash (or to_h) might work here.

spec/db/seeder_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+        name: Faker::Company.name,
+        storage_region: Fabricate(:storage_region).name,
+        storage_type: Fabricate(:storage_type).name,
+        fixity_algs: [Fabricate(:fixity_alg).name],
+        protocols: [Fabricate(:protocol).name],
+        api_root: Faker::Internet.url,
+        private_auth_token: Faker::Crypto.sha256,
+        auth_credential: Faker::Crypto.sha256
+      }
+    end
```
dazza-codes, line 153: Again, maybe a fabricator build with to_h might work here?

spec/db/seeder_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+      expect(created_node.protocols.pluck(:name)).to contain_exactly(*entry[:protocols])
+      expect(created_node.api_root).to eql(entry[:api_root])
+      expect(created_node.private_auth_token)
+        .to eql(Node.send(:generate_hash, entry[:private_auth_token]))
+      expect(created_node.auth_credential).to eql(entry[:auth_credential])
+    end
+    it "skips existing Node records" do
+      Fabricate(:node, namespace: entry[:namespace])
+      seeder.create(:nodes, [entry])
+      expect(logger.info_messages).to eql([skip_message])
```
darrenleeweber: info or warn message?

lib/db/seeder.rb
dazza-codes: BTW, also curious about whether the fabricator could be used in this class, maybe with a PORO generator (they have an example on their docs).

lib/db/seeder.rb
malakai97: * Re: putting the details into code directly: 
  Seeder already expects a ruby hash, not JSON.  Representing the data in a `.rb` file instead of .`json` would be fine, but doesn't gain us anything.

* Re: json-schema to validate the seeds
That would be more robust, but is overkill for a set of files that are changed so rarely.

* Re: ActiveRecord::Serialization
  The ruby builtin parser was sufficient for my needs.  The AR serializer expects json strings, meaning the seeder is dependent on the serialization format of the data files anyway.

spec/db/seeder_spec.rb
malakai97: On the fence about this.  The script is meant to be idempotent--e.g. a new member gets addded, run the entire thing.  So, skips will be normal.  I didn't want to issue a warning about normal behavior.

db/seeds.rb
dazza-codes: Oh, if it does not log anything, remove the logger from the class.  If it should own it's own logger, add it as a private method later, if it's required.

db/seeds/development-seeds.d/seeds.json
dazza-codes: OK, that's resolved this question (good to know that it's part of the spec).

lib/db/seed_file_loader.rb
dazza-codes: Add some specs to test what happens when given a non-existent directory.

lib/db/seeder.rb
dazza-codes: Let's consider how to skip the @map functionality - try saving a valid model.to_json string to a file and simply using that as the basis for the seeds data.  Then all this script has to do is read that file and pass it to model.from_json (that seems a lot simpler than all the mapping and other apparatus).  Also, if each file is named according to the model, it allows loading each of the named files in a sequence that will satisfy model dependencies.  At some level, all this script needs is an array of models to seed, in order.  It goes looking for a file corresponding to each model in the array.  Adding a new model could be as easy as adding it to the array, in the correct order to satisfy it's dependencies, and adding the corresponding json file.

dazza-codes (approved):
Hopefully the review comments spark some thoughts on revisions or suggest useful minor changes.  In general, there's nothing wrong, although it seems it could be simpler.  Leaving the details to Bryan, merge when ready or prompt me in slack for any additional input.

PR #144, opened 2016-12-05
Remove uniqueness validations on pseudo-primary keys (e.g. member_id)

### LINKED ISSUE #143: Duplicate ids raises RecordInvalid instead of RecordNotUnique
You can see this by creating a record with an <type>_id equal to one that already exists, regardless of other fields changing.

This is caused by the `uniqueness: true` validation, which causes the field's uniqueness to be validated before trying to save to the database.  Only the error originating in the database leads to a RecordNotUnique exception.

This validation is present on the <type>_id field of present on 
* Bag
* FixityCheck 
* Ingest
* Member
* Node
* ReplicationTransfer
* RestoreTransfer
* RunTime
* VersionFamily

Additionally, there is a compound constraint on MessageDigest via `validates_uniqueness_of :bag, scope: :fixity_alg`.  

Note that all of these constraints are already being enforced at the database level, should continue to be enforced there.

malakai97:
This resolves #143 

Following this change, we now leave it up to the database to detect a duplicate record (by our distributed primary key).  The biggest consequence of this is one must now attempt to save (or check via SELECT first) in order to know the record is duplicate.  

A few tests were altered to check uniqueness by checking for for the RecordNotUnique exception.

### DIFF
--- a/app/models/bag.rb
+++ b/app/models/bag.rb
@@ -62,7 +62,7 @@ class Bag < ActiveRecord::Base
   validates :version_family_id, read_only: true, on: :update
 
   ### Static Validations
-  validates :uuid, presence: true, uniqueness: true,
+  validates :uuid, presence: true,
             format: { with: /\A[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}\z/i,
             message: "must be a valid v4 uuid." }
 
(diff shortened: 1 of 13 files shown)


dazza-codes:
Codacy indicates potential new issues and take a quick look at the files tab and review any files that don't have an `A` rating to see if there's anything easy to resolve there.
- https://www.codacy.com/app/DPN/dpn-server/pullRequest?prid=441190

PR #145, opened 2016-12-05
Integration fixtures - replication_transfers - store_requested: true

dazza-codes:
Most of the replication transfers have `store_requested: false`, which seems to defeat the purpose of a replication transfer - at least it trips up some integration tests on the dpn-sync project.  It's seem more likely that these values should be `store_requested: true`, no?

### DIFF
--- a/test/fixtures/integration/aptrust/replication_transfers.yml
+++ b/test/fixtures/integration/aptrust/replication_transfers.yml
@@ -38,7 +38,7 @@ xfer01_to_sdr:
     to_node: sdr
     protocol: rsync
     fixity_alg: sha256
-    store_requested: false
+    store_requested: true
     stored: false
     cancelled: false
 
@@ -52,6 +52,6 @@ xfer01_to_tdr:
     to_node: tdr
     protocol: rsync
     fixity_alg: sha256
-    store_requested: false
+    store_requested: true
     stored: false
     cancelled: false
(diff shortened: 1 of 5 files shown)


malakai97:
`store_requested: true` indicates that the request has been accepted, the fixity value uploaded, and found to be correct by the from_node.  It seems to me that these fixtures are meant to represent "new" replication transfer requests.

dazza-codes:
Oh, where is the definition or intent of this field defined (apart from your memory)?  I had interpreted that field to mean - from_node is requesting the to_node to replicate a bag.  As you put it, it seems like the field name would be `store_accepted`?  Perhaps it depends on the perspective:  server vs. client.  Perhaps these field names need a `server` or `client` prefix?  Like `client_accepted: true || false` and `server_requested:  true || false`?  Then a client looks for `server_requested: true` to act on a replication request and a server looks for `client_accepted: true` as an indication that the replication is in process (or finished, depending on other fields).  So, looks like I need to do some homework on the field definitions and could use some advice on where to look for the details.

mikejritter:
The `store_requested` field comes from some of the api-v1 nomenclature, so it's something that's likely easy to miss. Basically, we needed a way to define the difference between transferring to a staging area and actually pushing a bag into the preservation repository. From this we had two status values: `received` (a node has transferred a bag) and `stored` (a node has stored a bag in its preservation area).

It's defined in swagger as:
```
store_requested (boolean): Set by the from_node after it receives a correct fixity_value to instruct the to_node to complete the replication
```
so it acts much in the same way as in api-v1 in that a node only sets the boolean after a bag has finished transferring and has had its fixity validated.

dazza-codes:
Closing this, as the PR is based on a misunderstanding of how the `store_requested` field is used.

PR #148, opened 2016-12-20
bundle update app_version_tasks

dazza-codes:
This new release of the app_version_tasks gem adds an option to push tags.  This update to the gem doesn't require any bump in the dpn-server version.


Pcolar:
OK - simple change - looks fine!

Issue #149, opened 2016-12-21
Replication controller occasionally returns a 422


mikejritter:
When completing a replication my client has seen the server return a 422 with the response body
```
{"status":"422","error":"Unprocessable Entity"}
```

According to TDR they saw an ActiveRecord exception in the log
```
ActiveRecord::RecordInvalid (Validation failed: Bag has already been taken):
 app/services/replication_transfer_updater.rb:137:in `add_replicating_node'
 app/services/replication_transfer_updater.rb:129:in `update'
 app/services/replication_transfer_updater.rb:10:in `update'
 app/controllers/replication_transfers_controller.rb:62:in `update'
```

There still needs to be some investigation as to what is going on, but at a first glance it seems like the error is coming from the replicating_nodes relationship. i.e. chron is already a replicating node of the bag, so an error is thrown when trying to add chron to the set. A good first step will be to look in to `app/services/replication_transfer_updater.rb`.

I should add that the replication itself is still updated successfully despite the 422 response.

mikejritter:
I was actually able to confirm this pretty easily using the rails console (truncated for brevity).

```
irb(main):003:0> b = Bag.find_by(uuid: "a32e18e4-8bf4-4bc0-9e98-3f51da9ea2a1")
irb(main):004:0>  n = Node.find_by(namespace: "chron")
irb(main):005:0> b.replicating_nodes << n
ActiveRecord::RecordInvalid: Validation failed: Bag has already been taken
```

I'm going to add a clause in the `replication_transfer_updater` and run a replication against it.

mikejritter:
Checking if the `replicating_nodes` includes a node before adding it seems to work well enough in my dev environment. Going to create a PR with the changes.

dazza-codes:
Confirmed a similar error on trying to update the `stored: true` attribute on a repl-transfer from TDR to SDR.  Somewhere in the TDR-dev node `dpn-server/log/production.log` there could be an error logged corresponding to the error logged in the SDR-dev node, i.e.
```
2017-01-07T20:00:34.955Z 16855 TID-o74f0 WARN: {"context":"Job raised exception”,  ... "created_at":1483819216.5947845,"enqueued_at":1483819216.5948524,"error_message":"Failed to update replication: {:status=>\"422\", :error=>\"Unprocessable Entity\”}"
```
[...]

PR #151, opened 2016-12-22
Replication Controller returns a 422

mikejritter:
From #149, this adds a check to make sure we don't add a duplicate entry to a nodes `replicating_nodes` field. When doing so the uniq constraint on the db table for the (node_id, bag_id) pair is violated, which causes rails to return a 422 Unprocessable Entity.

The only thing I can think of is if this is the best place for this or not. If the replicating_nodes field is not updated in any other location then I suppose it's fine, otherwise having it be in the bag model itself may be appropriate.

### DIFF
--- a/app/services/replication_transfer_updater.rb
+++ b/app/services/replication_transfer_updater.rb
@@ -134,7 +134,9 @@ class ReplicationTransferUpdater
     private
 
     def add_replicating_node
-      record.bag.replicating_nodes << record.to_node
+      unless record.bag.replicating_nodes.includes(record.to_node)
+        record.bag.replicating_nodes << record.to_node
+      end
     end
 
   end
--- a/app/services/replication_transfer_updater.rb
+++ b/app/services/replication_transfer_updater.rb
@@ -134,7 +134,7 @@ class ReplicationTransferUpdater
     private
 
     def add_replicating_node
-      unless record.bag.replicating_nodes.includes(record.to_node)
+      unless record.bag.replicating_nodes.include?(record.to_node)
         record.bag.replicating_nodes << record.to_node
       end
     end
--- a/spec/models/data_bag_spec.rb
+++ b/spec/models/data_bag_spec.rb
@@ -13,8 +13,8 @@ describe DataBag do
 
   it "has a factory that honors updated_at" do
     time = 1.year.ago
-    record = Fabricate(:data_bag, updated_at: 1.year.ago)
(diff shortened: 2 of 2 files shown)


dazza-codes:
Background note:  the SDR-dev node has encountered this bug (noted in #149) when doing a bag transfer.  The `dpn-sync` project does the bag transfer.  It attempts to update a replication-transfer twice during the bag transfer.  Once when the bag is transferred into the staging path (it updates the fixity_value and checks that the from_node has set the `store_requested: true` value).  It then attempts to transfer the bag to a storage facility and update the `stored: true` value on the from_node.  The 422 error was encountered during this second update.

dazza-codes:
One of the travis builds (for MySQL) has failed, i.e.
```
Failures:
  1) DataBag has a factory that honors updated_at

     Failure/Error: expect(record.updated_at.change(usec: 0)).to eql time.change(usec: 0)

       expected: 2016-01-06 18:48:50.000000000 +0000
            got: 2016-01-06 18:48:51.000000000 +0000
       (compared using eql?)

       Diff:
       @@ -1,2 +1,2 @@
       -Wed, 06 Jan 2016 18:48:50 UTC +00:00
       +Wed, 06 Jan 2016 18:48:51 UTC +00:00
     # ./spec/models/data_bag_spec.rb:17:in `block (2 levels) in <top (required)>'
```

This could be a symptom of a fragile rspec matcher.  It maybe better to use a `be_within` matcher that is not too picky about microseconds?  e.g.
https://www.relishapp.com/rspec/rspec-expectations/v/3-5/docs/built-in-matchers/be-within-matcher

dazza-codes (approved):
This looks like it should be OK, although please consider changing rspec matchers to use `be_within` (see conversation comments).  Once that is done and the travis builds are OK with it, we should be able to merge and ship this.

mikejritter:
Looking at the actual test it shows:
```
  it "has a factory that honors updated_at" do
    time = 1.year.ago
    record = Fabricate(:data_bag, updated_at: 1.year.ago)
    expect(record.updated_at.change(usec: 0)).to eql time.change(usec: 0)
  end 
```

While we should probably update this test to use time as an argument in the Fabricate, I don't think this PR is the place for that.

dazza-codes:
It's OK to make a change in this PR to the spec, e.g. try
```ruby
time = 1.year.ago
record = Fabricate(:data_bag, updated_at: time) # use time here?
expect(record.updated_at).to be_within(5).of(time.to_i) # or something like that?
```

dazza-codes (approved):

